PR #3, opened 2017-01-26
Add has_many through relationship to User and Bet models

ghost:
User model - name:string
Bet model - bet:text
BetManager - user_id:integer, bet_id: integer

### DIFF
--- a/app/models/bet_manager.rb
+++ b/app/models/bet_manager.rb
@@ -1,0 +1,4 @@
+class BetManager < ApplicationRecord
+  belongs_to :bet
+  belongs_to :user
+end
--- a/db/migrate/20170126214854_create_bet_managers.rb
+++ b/db/migrate/20170126214854_create_bet_managers.rb
@@ -1,0 +1,10 @@
+class CreateBetManagers < ActiveRecord::Migration[5.0]
+  def change
+    create_table :bet_managers do |t|
+      t.integer :user_id
+      t.integer :bet_id
+
+      t.timestamps
+    end
+  end
+end
--- a/app/models/bet.rb
+++ b/app/models/bet.rb
@@ -1,0 +1,4 @@
+class Bet < ApplicationRecord
+  has_many :users, through: :bet_manager
+  has_many :bet_manager
+end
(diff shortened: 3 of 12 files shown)


app/models/bet_manager.rb
```diff
@@ -0,0 +1,4 @@
+class BetManager < ApplicationRecord
```
qr8r, line 1: I am not a fan of `BetManager` as a name. It is a little bit nebulous as to what this model is actually responsible for. 

Everyone in the office has been linked to this article at somepoint: http://objology.blogspot.ca/2011/09/one-of-best-bits-of-programming-advice.html 

Given what this Class is doing I would lean more towards calling this a:  
Prediction
Outcome
UserBet (Kind of generic like UsersBets, but slighly better since its is actually what a `User` has `Bet` 
etc

db/migrate/20170126214854_create_bet_managers.rb
```diff
@@ -0,0 +1,10 @@
+class CreateBetManagers < ActiveRecord::Migration[5.0]
+  def change
+    create_table :bet_managers do |t|
+      t.integer :user_id
```
qr8r, line 4: This is correct, Rails also has the `#references` and `#belongs_to` migration methods which could be useful

Generally we include indexes and foreign key declarations when we are creating references.
https://tomafro.net/2009/08/using-indexes-in-rails-index-your-associations
https://robots.thoughtbot.com/referential-integrity-with-foreign-keys

qr8r (changes requested):
Excellent start! 

The content here is very good, but I'm going to rehash your fomatting a little: 

This should be 3 commits. One for each of the models you add to the database. Each of these commits should include the migration and schema changes.

Be careful when using generators, They may add a number of files you don't care about or want. In this case, the empty test files, and arguably the fixtures are not useful (and when we get to testing I'm going to ask you to use a different framework) 

The appdesign.md file is a useful document to have, I would split the changes to it into the 3 commits mentioned above.

Lastly, github is complaining that your branch cannot be merged due to a merge conflict. You will need to resolve the conflicts locally and push your branch up again before we can merge this into master

PR #4, opened 2017-01-27
User bet models

ghost:
Added migrations for User, Bet and Outcome models.

### DIFF
--- a/app/models/bet.rb
+++ b/app/models/bet.rb
@@ -1,0 +1,4 @@
+class Bet < ApplicationRecord
+  has_many :users, through: :outcome
+  has_many :outcome
+end
--- a/app/models/outcome.rb
+++ b/app/models/outcome.rb
@@ -1,0 +1,4 @@
+class Outcome < ApplicationRecord
+  belongs_to :bet
+  belongs_to :user
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,0 +1,4 @@
+class User < ApplicationRecord
+  has_many :bets, through: :outcome
+  has_many :outcome
+end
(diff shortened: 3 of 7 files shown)


qr8r (changes requested):
This commit is missing The associated schema.rb change 	43d8dca

db/schema.rb
```diff
@@ -10,7 +10,19 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170126234444) do
+ActiveRecord::Schema.define(version: 20170127000826) do
+
+  create_table "bets", force: :cascade do |t|
```
qr8r, line 20: this should be in the commit with Bets

db/migrate/20170126235850_add_bets.rb
```diff
@@ -0,0 +1,10 @@
+class AddBets < ActiveRecord::Migration[5.0]
```
qr8r, line 1: This commit should include The Bets model, and the schema changes

app/models/user.rb
```diff
@@ -1,4 +1,4 @@
 class User < ApplicationRecord
```
qr8r, line 1: This commit should be squashed into edd763668b20658fd95602ff812bebffcacabeb3

app/models/bet.rb
```diff
@@ -0,0 +1,4 @@
+class Bet < ApplicationRecord
```
qr8r, line 1: Squash into edd763668b20658fd95602ff812bebffcacabeb3

app/models/outcome.rb
```diff
@@ -0,0 +1,4 @@
+class Outcome < ApplicationRecord
```
qr8r, line 1: Squash into 3faaf8387b04310d14245578cb1f52fbcdc5fc29

qr8r (changes requested):

PR #5, opened 2017-01-27
Controllers branch

ghost:
This branch implements the controller and simple views for the User model.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,2 +1,5 @@
 class UsersController < ApplicationController
+  def index
+    @users = User.all
+  end
 end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,0 +1,2 @@
+class UsersController < ApplicationController
+end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -2,4 +2,8 @@ class UsersController < ApplicationController
   def index
     @users = User.all
   end
+
+  def new
+    @user = User.new
+  end
 end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -6,4 +6,18 @@ class UsersController < ApplicationController
   def new
     @user = User.new
   end
+
+  def create
+    @user = User.new(user_params)
+    if @user.save
+      redirect_to @user
+    else
+      render 'new'
+    end
+  end
+
+  private
+  def user_params
+    params.require(:user).permit(:username, :email, :password, :avatar)
+  end
 end
--- a/app/views/users/index.html.erb
+++ b/app/views/users/index.html.erb
@@ -1,0 +1,24 @@
+<table>
+  <thead>
+    <tr>
+      <th>
+        Username:
+      </th>
+      <th>
+        E-mail:
+      </th>
+    </tr>
+  </thead>
+  <tbody>
+    <% @users.each do |user| %>
+    <tr>
+      <td>
+        <%= user.username %>
+      </td>
+      <td>
+      <%= user.email %>
+      </td>
+    </tr>
+    <% end %>
+  </tbody>
+</table>
--- a/app/views/users/new.html.erb
+++ b/app/views/users/new.html.erb
@@ -1,0 +1,19 @@
+<%= form_for @user do |f| %>
+  <p>
+    <%= f.label :username %>
+    <%= f.text_field :username %>
+  </p>
+
+  <p>
+    <%= f.label :email %>
+    <%= f.text_field :email %>
+  </p>
+
+  <p>
+    <%= f.label :password %>
+    <%= f.text_field :password %>
+  </p>
+  <p>
+    <%= f.submit %>
+  </p>
+<% end %>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,3 +1,4 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+  resources :users
 end


qr8r (changes requested):
Needs a rebase and includes commits from master

config/routes.rb
```diff
@@ -1,3 +1,5 @@
 Rails.application.routes.draw do
```
qr8r, line 1: The commit message for this commit does not accurately describe the contents of the commit.

config/routes.rb
```diff
@@ -1,3 +1,5 @@
...
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+  root 'users#index'
```
qr8r: I think changing this route for testing is fine. But I can't really see why we would want the root of our website to be a list of users (im open to discussing this).

Likely we want to leave root as the default for now, until we create a PagesController to handle static pages like our homepage

config/routes.rb
```diff
@@ -1,5 +1,6 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
   root 'users#index'
+  get 'users/new'
```
qr8r: This is unneeded. The line below defines all of your RESTful routes for your user model.

To check which routes you have defined in your app you can run `bundle exec rake routes`

db/schema.rb
```diff
@@ -12,18 +12,18 @@
 
 ActiveRecord::Schema.define(version: 20170127000826) do
 
-  create_table "outcomes", force: :cascade do |t|
```
qr8r: This change is not pertinent to the feature you are implementing and probably doesnt need to be included in this commit

app/views/users/new.html.erb
```diff
@@ -0,0 +1,19 @@
+<%= form_for :user, url: users_path do |f| %>
```
qr8r: You can pass an actual model to the form builder and it will figure out which url to use based on the model you pass it 

`form_for @user do |f|`

app/views/users/index.html.erb
```diff
@@ -0,0 +1,20 @@
...
+      Username:
+    </th>
+    <th>
+      Password:
+    </th>
+  </tr>
+<% @users.each do |user| %>
+  <tr>
+    <td>
+      <%= render plain: user.username %>
```
qr8r: You only need render if you are interpreting the contents of a template.

`<%= user.username %>` should be sufficient here

app/views/users/index.html.erb
```diff
@@ -0,0 +1,20 @@
...
+    </td>
+    <td>
+    <%= render plain: user.email %>
```
qr8r: ^^

app/views/users/index.html.erb
```diff
@@ -0,0 +1,20 @@
+<table>
+  <tr>
+    <th>
+      Username:
+    </th>
+    <th>
+      Password:
+    </th>
+  </tr>
+<% @users.each do |user| %>
```
qr8r: Excellent use of looping here! 

Be careful of your indentation. This loop and its contents should be 2 spaces indented from the `<table>` tag

app/views/users/index.html.erb
```diff
@@ -0,0 +1,20 @@
+<table>
+  <tr>
```
qr8r: If you wanted to be fancy you could use the `<thead>` and `<tbody>` HTML5 tags

qr8r (changes requested):
There is no UsersController included in this PR

app/views/users/new.html.erb
```diff
@@ -0,0 +1,19 @@
+<%= form_for :user, url: @user do |f| %>
```
qr8r: This can just be `<%= form_for @user do |f| %>`

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,23 @@
+class UsersController < ApplicationController
```
qr8r, line 1: There are no blank views included in this commit

config/routes.rb
```diff
@@ -1,5 +1,4 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
-  root 'users#index'
```
qr8r: This change to `root` occured in the previous commit. Can you squash this removal into the previous commit?

db/schema.rb
```diff
@@ -12,11 +12,6 @@
...
```
qr8r: Im still pretty sure you didnt mean to remove the `outcomes` table from your database schema here. You might want to kill this removal from this commit

qr8r (changes requested):
Cool! Looking really goo and this definitely works. Try and be more aware of how you are describing a commit (with the commit message) and what code you are actually putting into the commit. If a chunk of code doesn't quite fit with the contents of the commit you're working on, it's totally cool to break it into 2 commits. 

Remember to add a PR description please :)

config/routes.rb
```diff
@@ -1,3 +0,0 @@
-Rails.application.routes.draw do
```
qr8r: Why are you deleting the routes file?

qr8r (changes requested):
Really great job taking the feedback about smaller commits to heart. This PR is a lot easier to read now. 

Be careful not to break commits into such small chunks that the code no longer works.

Commits should be the smallest possible chunk of code *that is still functional*

for example. Adding the controller full of blank actions with out the associated views wont work. If you attempt to visit any of the pages you will get a template not found error.

Another example of over splitting code is separating the `user_params` strong params guard from the `create` action where it is actually used. In this case, at the point when the `user_params` method is added, it is never used anywhere, this is what we call "zombie code" code that exists but never runs.

Your commits are definitely way better, it just takes time to understand what constitutes the "smallest possible coherent and functional commit"

qr8r (approved):

PR #6, opened 2017-01-27
User bet models

ghost:


db/migrate/20170127000826_add_outcomes.rb
```diff
@@ -0,0 +1,8 @@
+class AddOutcomes < ActiveRecord::Migration[5.0]
+  def change
+    create_table :outcomes do |t|
+      t.integer :user_id
```
qr8r, line 7: References to other database tables should be indexed and have a foreign key 

https://tomafro.net/2009/08/using-indexes-in-rails-index-your-associations
https://robots.thoughtbot.com/referential-integrity-with-foreign-keys

qr8r (changes requested):

qr8r (approved):

PR #9, opened 2017-02-03
Defining Essential Models and their Relationships

ghost:
Established associations between User, Bet and Option models through UserBets join table.

### DIFF
--- a/app/models/outcome.rb
+++ b/app/models/outcome.rb
@@ -1,4 +1,0 @@
-class Outcome < ApplicationRecord
-  belongs_to :bet
-  belongs_to :user
-end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,4 +1,4 @@
 class User < ApplicationRecord
-  has_many :bets, through: :outcome
-  has_many :outcome
+  has_many :bets, through: :user_bets
+  has_many :user_bets
 end
--- a/app/models/user_bet.rb
+++ b/app/models/user_bet.rb
@@ -1,0 +1,4 @@
+class UserBet < ApplicationRecord
+  belongs_to :user
+  belongs_to :bet
+end
--- a/app/models/user_bet.rb
+++ b/app/models/user_bet.rb
@@ -1,4 +1,5 @@
 class UserBet < ApplicationRecord
   belongs_to :user
   belongs_to :bet
+  belongs_to :bet_option
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170202212027) do
+ActiveRecord::Schema.define(version: 20170202234915) do
 
   create_table "bet_options", force: :cascade do |t|
     t.text     "option_text"
@@ -23,6 +23,7 @@ ActiveRecord::Schema.define(version: 20170202212027) do
     t.integer  "status"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.integer  "creator_id"
   end
 
   create_table "user_bets", force: :cascade do |t|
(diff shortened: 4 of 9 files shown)


db/migrate/20170202234757_add_credits_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddCreditsToUsers < ActiveRecord::Migration[5.0]
+  def change
+    add_column :users, :credits, :integer
```
qr8r: For now lets assume that all users have the money to back up bets. The system doesnt need to track how many credits a user has

db/migrate/20170127000826_add_outcomes.rb
```diff
@@ -1,11 +0,0 @@
-class AddOutcomes < ActiveRecord::Migration[5.0]
```
qr8r: Dont delete existing migrations. To change existing structure, add new migrations

app/models/outcome.rb
```diff
@@ -1,4 +0,0 @@
-class Outcome < ApplicationRecord
-  belongs_to :bet
```
qr8r, line 2: Doesnt belong in this commit

app/models/user.rb
```diff
@@ -1,4 +1,6 @@
 class User < ApplicationRecord
-  has_many :bets, through: :outcome
-  has_many :outcome
+  has_many :bets, through: :user_bets
+  has_many :user_bets
+
+  has_secure_password
```
qr8r: From the session branch

db/migrate/20170202232146_add_option_total_to_options.rb
```diff
@@ -0,0 +1,5 @@
+class AddOptionTotalToOptions < ActiveRecord::Migration[5.0]
+  def change
```
qr8r: These migrations add colums to the same model in the same commit. It can probably be the same migration

app/models/option.rb
```diff
@@ -0,0 +1,4 @@
+class Option < ApplicationRecord
```
qr8r: Option is a very generic name for this. Now that we have changed what used to be `Outcomes` to `UserBet` I think outcome more accurately describes what this model represents

db/schema.rb
```diff
@@ -20,6 +20,13 @@
     t.integer  "creator_id"
   end
 
+  create_table "options", force: :cascade do |t|
+    t.datetime "created_at",   null: false
+    t.datetime "updated_at",   null: false
+    t.text     "option_text"
+    t.integer  "option_total"
```
qr8r: So every user who bets on the same option (outcome) has to bet the same amount of money ?

app/models/user_bet.rb
```diff
@@ -0,0 +1,5 @@
+class UserBet < ApplicationRecord
+  belongs_to :user
```
qr8r, line 2: Should be in the same commit as when you remove the Outcomes model earlier

db/migrate/20170202212027_create_user_bets.rb
```diff
@@ -0,0 +1,8 @@
+class CreateUserBets < ActiveRecord::Migration[5.0]
+  def change
```
qr8r: all of these migrations should be merged into a single migration

db/migrate/20170202212555_add_user_to_user_bets.rb
```diff
@@ -0,0 +1,5 @@
+class AddUserToUserBets < ActiveRecord::Migration[5.0]
+  def change
+    add_reference :user_bets, :user, foreign_key: true
```
qr8r: dont forget to index references as well

qr8r (changes requested):
These commits are well sized but they are not coherent. Some commits contain code they shouldn't need.

Another rule of thumb is that the project should be 'functional' after every commit. So that I could checkout any commit you make and the project would still run without errors

qr8r (approved):

PR #11, opened 2017-02-06
Adding Debt model

ghost:
-Created Debt model with the following columns:
  * debtor_id:   user who owes money;
  * creditor_id: user who debtor owes to;
  * amount:      decimal number with two point precision.
-Added belongs_to associations with User (under debtor and creditor) to Debt;
-Added has_many association with Debt to User.

### DIFF
--- a/app/models/debt.rb
+++ b/app/models/debt.rb
@@ -1,0 +1,4 @@
+class Debt < ApplicationRecord
+  belongs_to :debtor,   class_name: 'User'
+  belongs_to :creditor, class_name: 'User'
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,4 +1,6 @@
 class User < ApplicationRecord
   has_many :bets, through: :user_bets
   has_many :user_bets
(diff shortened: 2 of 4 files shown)


app/models/debt.rb
```diff
@@ -0,0 +1,4 @@
+class Debt < ApplicationRecord
+  belongs_to :debtor,   class_name: 'User', foreign_key: :debtor_id
```
qr8r: Are you sure the foreign key needs to be specifed here?

app/models/user.rb
```diff
@@ -1,4 +1,5 @@
 class User < ApplicationRecord
   has_many :bets, through: :user_bets
   has_many :user_bets
+  has_many :debts, foreign_key: :debtor_id
```
qr8r: Can a user also have credits? ie

```ruby
user.debts # => Collection of debts user owes to other people
user.credits #=> Collection of debts owed to user 
```

qr8r (approved):

PR #12, opened 2017-02-06
Add references to UserBet model

ghost:
-Added references to UserBet model (instead of previous integer columns);
-Added win_option boolean column to BetOption model that lets it know if this particular option is a winner or not.

### DIFF
--- a/db/migrate/20170207174141_add_bet_option_to_user_bets.rb
+++ b/db/migrate/20170207174141_add_bet_option_to_user_bets.rb
@@ -1,0 +1,7 @@
+class AddBetOptionToUserBets < ActiveRecord::Migration[5.0]
+  def change
+    add_column      :user_bets, :bet_option_id, :integer
+    add_foreign_key :user_bets, :bet_options
+    add_index       :user_bets, :bet_option_id
+  end
+end
--- a/app/models/bet.rb
+++ b/app/models/bet.rb
@@ -1,5 +1,5 @@
 class Bet < ApplicationRecord
   has_many :users, through: :user_bets
   has_many :user_bets
-  has_many :options, through: :user_bets
+  has_many :options, through: :user_bets, source: :bet_option
 end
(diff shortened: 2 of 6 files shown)


db/migrate/20170206211813_add_references_to_user_bet.rb
```diff
@@ -0,0 +1,11 @@
+class AddReferencesToUserBet < ActiveRecord::Migration[5.0]
+  def change
+    drop_table :user_bets
```
qr8r: Consider why you wouldnt want to do this on a live production database

db/migrate/20170206222252_add_win_option_to_bet_options.rb
```diff
@@ -0,0 +1,5 @@
+class AddWinOptionToBetOptions < ActiveRecord::Migration[5.0]
+  def change
+    add_column(:bet_options, :win_option, :boolean)
```
qr8r: I would prefer `:winner` over `:win_option`. Might just be a personal preference though

qr8r (changes requested):

db/migrate/20170206211813_add_references_to_user_bet.rb
```diff
@@ -0,0 +1,12 @@
+class AddReferencesToUserBet < ActiveRecord::Migration[5.0]
+  def change
+    change_table :user_bets do |t|
```
qr8r: This is also bad. 

What effect would this have if it were run on a production database? 
What is the goal of the migration? (what are you trying to accomplish) 
Can you achieve that goal in a less destructive way?

qr8r (changes requested):

db/migrate/20170207174141_add_bet_option_to_user_bets.rb
```diff
@@ -0,0 +1,7 @@
+class AddBetOptionToUserBets < ActiveRecord::Migration[5.0]
+  def change
+    add_column      :user_bets, :bet_option, :integer
```
qr8r: Remember what we talked about yesterday, and make sure that the core you run actually does what you expect. 

This could also have been done in one step: http://apidock.com/rails/v4.0.2/ActiveRecord/ConnectionAdapters/SchemaStatements/add_reference 

Theres nothing wrong with doing it in multiple steps however I dont think this give you what you actually want

qr8r (changes requested):

qr8r (approved):

PR #13, opened 2017-02-07
Resolving Bets

ghost:
The resolve method creates an entry in the Debt model that indicates who owes who and how much based on the amount bet by both losing and winning sides. 
Owing amount is calculated individually for each bet loser.


app/models/bet.rb
```diff
@@ -2,4 +2,42 @@ class Bet < ApplicationRecord
...
+      debt = UserBet.find_by(user: debtor, bet: self).amount_bet
+      win_contributions.each do |contribution|
+        Debt.create(debtor: debtor,
+                    creditor: contribution[:creditor],
+                    amount:   debt * contribution[:percentage])
+      end
+    end
+  end
+
+  def win_option
```
qr8r: Should be a scope on Option

http://api.rubyonrails.org/classes/ActiveRecord/Scoping/Named/ClassMethods.html#method-i-scope

app/models/bet.rb
```diff
@@ -2,4 +2,42 @@ class Bet < ApplicationRecord
...
+    options.find(&:winner)
+  end
+
+  def winner_total
+    UserBet.where(bet: self, bet_option: win_option).map(&:amount_bet).sum
+  end
+
+  def winners
```
qr8r: Could maybe be simplified with a scope on UserBet

app/models/bet.rb
```diff
@@ -2,4 +2,42 @@ class Bet < ApplicationRecord
...
+    UserBet.where(bet: self, bet_option: win_option).map do |ub|
+      User.find(ub.user_id)
+    end
+  end
+
+  def losers
```
qr8r: Could maybe be simplified with a scope on UserBet

app/models/bet.rb
```diff
@@ -2,4 +2,42 @@ class Bet < ApplicationRecord
...
+    UserBet.where(bet: self).where.not(bet_option: win_option).map do |ub|
+      User.find(ub.user_id)
+    end
+  end
+
+  def win_contributions
```
qr8r: Could be probably also be simplified with a scope on UserBet.

Returning an array of hashes seems a little weird to me here

app/models/bet.rb
```diff
@@ -2,4 +2,42 @@ class Bet < ApplicationRecord
   has_many :users, through: :user_bets
   has_many :user_bets
   has_many :options, through: :user_bets, source: :bet_option
+
+  def resolve
+    losers.each do |debtor|
+      debt = UserBet.find_by(user: debtor, bet: self).amount_bet
+      win_contributions.each do |contribution|
+        Debt.create(debtor: debtor,
```
qr8r: This can be build off of a users. `user.debts.create(amount: x. creditor: y)`

qr8r (commented):
Nice job adding methods 1 commit at a time! These are nice and small and easy to digest. 
I suggest looking into the concept of an ActiveRecord scope, it will help you simplify how you are looking up records from the database.

In general good job naming your methods, they all seem to have simple names which describe what the code is actually doing (semantic).

app/models/user_bet.rb
```diff
@@ -2,4 +2,10 @@ class UserBet < ApplicationRecord
   belongs_to :user
   belongs_to :bet
   belongs_to :bet_option
+
+  scope :winner_total, lambda { |win_opt|
+    where(bet_option: win_opt)
+      .map(&:amount_bet)
```
qr8r: You generally want to avoid performing ruby calculations in scopes. We use scopes primarily to simplify our interactions with the datable and between models

app/models/bet.rb
```diff
@@ -2,4 +2,41 @@ class Bet < ApplicationRecord
...
+
+  def win_option
+    options.winners
+  end
+
+  def winner_total
+    user_bets.winners.map(&:amount_bet).sum
+  end
+
+  def winners
```
qr8r: I dont think we need a helper method for this. Being able to type `winners` is not much different than typing `users.winners` in the code when you need it

app/models/bet.rb
```diff
@@ -2,4 +2,41 @@ class Bet < ApplicationRecord
...
+    users.winners
+  end
+
+  def losers
+    users.losers
```
qr8r: Uneeded helper method

app/models/bet_option.rb
```diff
@@ -1,4 +1,7 @@
 class BetOption < ApplicationRecord
   has_many :users, through: :user_bets
   has_many :user_bets
+
+  scope :winners, -> { where(winner: true) }
+  scope :losers, -> { where(winner: [nil, false]) }
```
qr8r: Since this is a boolean field, lets make the db column default to false. This should usually be done for any boolean db column. That way your db column is guaranteed to be either true or false. ( also probably want to require this column and not allow it to be blank)

app/models/user.rb
```diff
@@ -3,4 +3,14 @@ class User < ApplicationRecord
   has_many :user_bets
   has_many :debts,   class_name: 'Debt', foreign_key: :debtor_id
   has_many :credits, class_name: 'Debt', foreign_key: :creditor_id
+
+  scope :winners, lambda {
```
qr8r: Curly braces are for single line blocks.  Multiline block require `do...end` 
also note that you will have to get fancy with parenthesis by moving to the do/end syntax.

for lambdas preferred syntax is: 
oneline:
`->{...}`

multiline:
```ruby
lambda do
...
end
```

app/models/user.rb
```diff
@@ -3,4 +3,14 @@ class User < ApplicationRecord
...
+    joins(user_bets: :bet_option)
+      .merge(BetOption.winners)
+  }
+
+  scope :losers, lambda {
+    joins(user_bets: :bet_option)
+      .merge(BetOption.losers)
```
qr8r: Javascript prefers dots at the start of the line 
Ruby usually prefers dot at the end of the previous line (this is an opinion) but reading line by line indicates that there is a continuation on the next line if the dot is at the end, and the code can be copy and pasted into a terminal.

If the dot is at the start of the new line, the code can't be copy pasted directly into a terminal

app/models/user_bet.rb
```diff
@@ -2,4 +2,6 @@ class UserBet < ApplicationRecord
   belongs_to :user
   belongs_to :bet
   belongs_to :bet_option
+
+  scope :winners, -> { joins([:user, :bet_option]).merge(BetOption.winners) }
```
qr8r: Un-needed array here. `joins` accepts a splat, which means you can pass in aguments directly without wrapping them in an array:

`joins(:user, :bet_option)` 

http://jacopretorius.net/2012/01/splat-operator-in-ruby.html

app/models/bet.rb
```diff
@@ -2,4 +2,41 @@ class Bet < ApplicationRecord
...
+  end
+
+  def amount_bet(user)
+    user_bets.find_by(user: user).amount_bet
+  end
+
+  def win_contributions
+    winners.map do |w|
+      winner_individual = amount_bet(w)
+      { creditor: w, percentage: (winner_individual / winner_total).round(2) }
```
qr8r: This method returns a complex data structure of the form:

```ruby
[
  { creditor:, percentage: },
  { creditor:, percentage: },
  { creditor:, percentage: }
]
```

This feels overly complex and I think you can achieve a similar effect more simply. If you really needed to return such a complicated data structure, it would probably be an indication that you are missing a class to model that data.

app/models/bet.rb
```diff
@@ -2,4 +2,41 @@ class Bet < ApplicationRecord
   has_many :users, through: :user_bets
   has_many :user_bets
   has_many :options, through: :user_bets, source: :bet_option
+
+  def resolve
+    losers.each do |debtor|
+      debt = amount_bet(debtor)
+      win_contributions.each do |contribution|
+        debtor.debts.create(creditor: contribution[:creditor],
```
qr8r: Indentation matters for readability...

This is something I should make sure rubocop is catching but accepted indentations are: 

```ruby
# in this indentation the closing parens lines up with the starting line of the function giving a visual 
# indication of what is encapsulated in the method call
debtor.debts.create(
  creditor: contribution[:creditor],
  amount:   debt * contribution[:percentage]
)
```

```ruby
# in this indentation the parens of the method being called line up indicating which arguments are being passed to a 
# particular method 
debtor.debts.create(
                     creditor: contribution[:creditor],
                     amount:   debt * contribution[:percentage]
                   )
```

qr8r: Another Comment on this method is What does it return. 

When designing methods you should try and think about them as atomic black boxes; code outside the method cannot see what is happening inside the method (even though they are not). Think:
- What information needs to go in
- What information needs to come out

We try and avoid side effects when writing methods. A side effect is when a method changes something in the system  with no indication externally that something has changed. In the case of this method. There is no indication outside of this method that debts are created.

Side effects make code hard to understand and hard to test. If I have a method:

```ruby
def add(num1, num2
  puts "#{num1 + num2}"
end

# and 

def add(num1, num2)
  num1 + num2
end
```

The second method is easier to test. I would write something like:
`expect(add(1,2)).to eq 3` 

But for the first one, the addition happens and the result is printed to the screen as a side effect. Much harder to test.

Think about that this method doing and what it should return in order to reduce side-effects happening within the blackbox of the method.

app/models/bet.rb
```diff
@@ -2,4 +2,41 @@ class Bet < ApplicationRecord
...
+
+  def winners
+    users.winners
+  end
+
+  def losers
+    users.losers
+  end
+
+  def amount_bet(user)
```
qr8r: I think this is a really good method. It is semantic (the name indicates what it does) it takes a reasonable argument and it does 1 thing and returns the output of that one thing. 

I am not sure however that this is functionality that is needed. I think a different approach would be to iterate over the user bets without caring specifically which user they belong to

qr8r (changes requested):

qr8r:
@YarekTenko This branch needs a rebase

app/models/bet_option.rb
```diff
@@ -3,4 +3,5 @@ class BetOption < ApplicationRecord
   has_many :user_bets
 
   scope :winners, -> { where(winner: true) }
+  scope :losers, -> { where(winner: false) }
```
qr8r: Technically this commit needs to be earlier in the history for all of your commits to work. But dont worry about fixing it.

In one of your previous commits you try and use this scope before you define it

qr8r (approved):

PR #16, opened 2017-02-10
Add spec for Bet model

ghost:
Tests winner_total, options.winners and resolve methods for Bet model.

### DIFF
--- a/app/models/bet.rb
+++ b/app/models/bet.rb
@@ -3,7 +3,8 @@ class Bet < ApplicationRecord
   has_many :user_bets
   has_many :options, through: :user_bets, source: :bet_option
 
-  def resolve
+  def resolve(winning_option:)
+    winning_option.update!(winner: true)
     debt_relationships = user_bets.winners.to_a.product(user_bets.losers.to_a)
     debt_relationships.map do |(won_bet, lost_bet)|
       Debt.create(
--- a/spec/models/bet_spec.rb
+++ b/spec/models/bet_spec.rb
@@ -1,0 +1,63 @@
+require 'rails_helper'
+
+RSpec.describe Bet, type: :model do
+  let!(:bob) do
+    User.create(username: 'Bob', email: 'bob@mail.com', password: '123')
+  end
+  let!(:peter) do
+    User.create(username: 'Peter', email: 'peter@mail.com', password: 'pete1')
+  end
+  let!(:jack) do
+    User.create(username: 'Jack', email: 'jack@mail.com', password: 'notjack')
+  end
+
+  let(:bet1) do
+    Bet.create(bet: 'Superbowl: Atlanta vs New England, place your bets!')
+  end
+  let(:option1) { BetOption.create(option_text: 'Atlanta wins') }
+  let(:option2) { BetOption.create(option_text: 'New England wins') }
+
+  let!(:users_place_bets) do
+    UserBet.create(user:       bob,
+                   bet:        bet1,
+                   bet_option: option1,
+                   amount_bet: 100)
+
+    UserBet.create(user:       peter,
+                   bet:        bet1,
+                   bet_option: option1,
+                   amount_bet: 50)
+
+    UserBet.create(user:       jack,
+                   bet:        bet1,
+                   bet_option: option2,
+                   amount_bet: 300)
+  end
+
+  describe('#resolve') do
+    subject(:debts) { bet1.resolve(winning_option: option) }
+    let(:option) { option1 }
+    it 'is an array of debts' do
+      expect(debts).to all be_a Debt
+    end
+
+    context 'Atlanta wins' do
+      let(:expected_debt) do
+        Debt.find_by!(debtor: jack, creditor: bob, amount: 201)
+      end
+      it 'contains the correct debts' do
+        expect(debts).to include expected_debt
+      end
+    end
+
+    context 'New England wins' do
+      let(:expected_debt) do
+        Debt.find_by!(debtor: bob, creditor: jack, amount: 100)
+      end
+      let(:option) { option2 }
+      it 'contains the correct debts' do
+        expect(debts).to include expected_debt
+      end
+    end
+  end
+end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -14,3 +14,12 @@ Style/FrozenStringLiteralComment:
 
 Style/Documentation:
   Enabled: false
+
+Metrics/BlockLength:
+  ExcludedMethods:
+    - 'describe'
+    - 'context'
+
+Metrics/LineLength:
+  IgnoredPatterns:
+    - '\w*it.+do'


qr8r:
@YarekTenko I will fix the rubo cop for this one

spec/models/bet_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+                   amount_bet: 300)
+  end
+
+  let(:ub) do
+    ub1
+    ub2
+    ub3
+  end
+
+  context_string = 'Bet: Superbowl - Atlanta vs New England;\n' \
```
qr8r: Ive changed the rubocop so that specs wont care if your lines are long. 

In specs its more important to write them so that they are readable and helpful, so writing long context statements is allows. 

This is one of the few cases where saving this string to a variable is bad because it hides the content which should be very obvious for future devs reading this file.

spec/models/bet_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+    'Betters on Atlanta: Bob ($100), Peter ($50);\n' \
+    'Betters on New England: Jack ($300).'
+
+  context context_string do
```
qr8r: This context doesnt add a whole lot of value.  generally a `context` statement is used to change the test setup for specific edge cases:

```ruby
describe '#foo' do
  context 'on tuesdays this returns true' do
    let(:date) { :tuesday } 
    it { is_expected.to be truthy }
  end 
  context 'on mondays this returns false' do
    let(:date) { :monday }
    it { is_expected.to be truthy }
  end
end
```
This context is acting more as a description than changing the setup behaviour. If you feel your test needs to be more clear about what it doing, try and write your `let` statements so they are more descriptive about what they return

spec/models/bet_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+  end
+
+  let(:ub2) do
+    UserBet.create(user:       peter,
+                   bet:        bet1,
+                   bet_option: option1,
+                   amount_bet: 50)
+  end
+
+  let(:ub3) do
```
qr8r: This is a bad let statement ub3 is not very descriptive

spec/models/bet_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+  context_string = 'Bet: Superbowl - Atlanta vs New England;\n' \
+    'Betters on Atlanta: Bob ($100), Peter ($50);\n' \
+    'Betters on New England: Jack ($300).'
+
+  context context_string do
+    describe('#winner_total') do
+      subject { bet1.winner_total }
+      context 'Atlanta wins' do
+        it 'returns the total amount of money bet on Atlanta winning (150)' do
+          ub
```
qr8r: Look up the difference between `let` and `let!`

spec/models/bet_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+          option1.update(winner: true)
+          expect(subject).to eq 150
```
qr8r: This is what we call a "magic number" we try and avoid having numbers floating around in code without any context, because its very easy to become confused about what they represent.

This should be encapsulated in a `let` statement ..... something like `let(:expected_debt_amount) { ... }

spec/models/bet_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+        end
+      end
+    end
+
+    describe('#options.winners') do
+      subject { bet1.options.winners }
+      context 'Atlanta wins' do
+        it 'returns "Atlanta Wins" option as the winning option' do
+          ub
+          option1.update(winner: true)
```
qr8r: Why are you changing the object you're testing from within your test.

spec/models/bet_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+
+    describe('#resolve') do
+      subject { bet.resolve }
+      context 'Atlanta wins' do
+        it 'creates entries in Debt model so that: Jack owes Bob 201; Jack owes Peter 99' do
+          ub
+          option1.update(winner: true)
+          bet1.resolve
+          expect(subject.find_by(debtor: jack, creditor: bob).amount).to eq 201
+          expect(subject.find_by(debtor: jack, creditor: peter).amount).to eq 99
```
qr8r: As much as possible try and avoid multiple expectations per test. 

sometimes its failures and then you will want to use the `:aggregate_failures` feature of rspec

qr8r (changes requested):

spec/models/bet_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+                   bet_option: option2,
+                   amount_bet: 300)
+  end
+
+  describe('#winner_total') do
+    subject { bet1.winner_total }
+    context 'Atlanta wins' do
+      let(:expected_atlanta_win_total) { 150 }
+      it 'returns the total amount of money bet on Atlanta winning (150)' do
+        users_place_bets
```
qr8r: Not needed because you define this in a `let!`

spec/models/bet_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+                   bet:        bet1,
+                   bet_option: option2,
+                   amount_bet: 300)
+  end
+
+  describe('#winner_total') do
+    subject { bet1.winner_total }
+    context 'Atlanta wins' do
+      let(:expected_atlanta_win_total) { 150 }
+      it 'returns the total amount of money bet on Atlanta winning (150)' do
```
qr8r: The number should be removed here. It could be changed and then we have to update it multiple places. 

`it 'is the correct money bet amount' ` would be enough information

spec/models/bet_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+      end
+    end
+  end
+
+  describe('#resolve') do
+    subject { bet1.resolve }
+    context 'Always' do
+      it 'returns array of newly created debts' do
+        users_place_bets
+        option1.update(winner: true)
```
qr8r: This should be part of the setup that is relevant to this context block

spec/models/bet_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+        expect(subject).to eq Debt.all
+      end
+    end
+    context 'Atlanta wins' do
+      let(:expected_debt_jack_to_bob) { 201 }
+      let(:actual_debt_jack_to_bob) do
+        Debt.find_by(debtor: jack, creditor: bob).amount
```
qr8r: You should be able to reference values defined in previous let blocks

qr8r (changes requested):
Ping me tomorrow before you start working and we will go over this together :)

app/models/bet.rb
```diff
@@ -3,7 +3,8 @@ class Bet < ApplicationRecord
   has_many :user_bets
   has_many :options, through: :user_bets, source: :bet_option
 
-  def resolve
+  def resolve(winning_option:)
+    winning_option.update(winner: true)
```
qr8r: Whats the difference between `update` and `update!` ?

spec/models/bet_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    let(:option) { option1 }
+    it 'is an array of debts' do
+      expect(debts).to all be_a Debt
+    end
+
+    context 'Atlanta wins' do
+      let(:expected_debt) do
+        Debt.find_by!(debtor: jack, creditor: bob, amount: 201)
+      end
+      let(:option) { option1 }
```
qr8r: Technically this line is not needed  since it is the same as the let statement defined on line 38

qr8r (dismissed):
This is 99% there. Just a couple super minor comments and then Im happy to approve this guy

qr8r (approved):

PR #18, opened 2017-02-14
Add password_digest column to User

ghost:
Added has_secure_password method that decrypts the provided password
(with bcrypt) and stores it in the password_digest column.


qr8r:
This should be part of the #17 PR Both are very good commits, but they are both related to the same functionality. Ill approve both, but in future if 2 commits are strongly dependant on eachother (ie, has secure password does not work at all without bcrpt) Then they should probably be part of the same PR

qr8r (approved):

PR #19, opened 2017-02-14
Add PagesController

ghost:
Pointed root to pages#index.

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -1,0 +1,3 @@
+class PagesController < ApplicationController
+  def index; end
+end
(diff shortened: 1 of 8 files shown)


app/controllers/pages_controller.rb
```diff
@@ -0,0 +1,3 @@
+class PagesController < ApplicationController
+  def index; end
```
qr8r, line 2: Fun Fact: this is technically not needed. but being explicit and not relying on too much rails magic is good.

I just wanted to point out that rails will figure out that it needs to render the index.html.erb file if there is no index action defined

qr8r (approved):

PR #22, opened 2017-02-15
Implement basic session control

ghost:
Add functionality to log in and log out as a user.

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -1,7 +1,15 @@
 class SessionsController < ApplicationController
   def new; end
 
-  def create; end
+  def create
+    user = User.find_by(email: params[:session][:email])
+    if user && user.authenticate(params[:session][:password])
+      log_in user
+      redirect_to user
+    else
+      render 'new'
+    end
+  end
 
   def destroy; end
 end
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -11,5 +11,8 @@ class SessionsController < ApplicationController
     end
   end
 
-  def destroy; end
+  def destroy
+    log_out
+    redirect_to root_path
+  end
 end
--- a/app/helpers/sessions_helper.rb
+++ b/app/helpers/sessions_helper.rb
@@ -1,0 +1,5 @@
+module SessionsHelper
+  def log_in(user)
+    session[:user_id] = user.id
+  end
+end
--- a/app/helpers/sessions_helper.rb
+++ b/app/helpers/sessions_helper.rb
@@ -3,6 +3,11 @@ module SessionsHelper
     session[:user_id] = user.id
   end
 
+  def log_out
+    session.delete(:user_id)
+    @current_user = nil
+  end
+
   def current_user
     @current_user ||= User.find_by(id: session[:user_id])
   end
--- a/app/helpers/sessions_helper.rb
+++ b/app/helpers/sessions_helper.rb
@@ -2,4 +2,8 @@ module SessionsHelper
   def log_in(user)
     session[:user_id] = user.id
   end
+
+  def current_user
+    @current_user ||= User.find_by(id: session[:user_id])
+  end
 end
--- a/app/views/pages/index.html.erb
+++ b/app/views/pages/index.html.erb
@@ -6,5 +6,6 @@
     <hr>
     <%= current_user.username %><br>
     <%= current_user.email %>
+    <%= link_to 'log out', session_path, method: :delete %>
   <% end %>
 </div>
--- a/app/views/pages/index.html.erb
+++ b/app/views/pages/index.html.erb
@@ -7,5 +7,15 @@
     <%= current_user.username %><br>
     <%= current_user.email %>
     <%= link_to 'log out', session_path, method: :delete %>
+  <% else %>
+    <h2>Welcome, stranger!</h2><br>
+      <div class="row">
+        <div class="col col-lg-1 offset-lg-5">
+          <%= link_to 'Log in', new_session_path %>
+        </div>
+        <div class="col col-lg-1">
+          <%= link_to 'Sign up', new_user_path %>
+        </div>
+      </div>
   <% end %>
 </div>
--- a/app/views/pages/index.html.erb
+++ b/app/views/pages/index.html.erb
@@ -1,4 +1,10 @@
 <div class="container text-center">
   <h1>Pages#index</h1>
   <p>Find me in app/views/pages/index.html.erb</p>
+  <% if current_user %>
+    <h2>Logged-in User:</h2>
+    <hr>
+    <%= current_user.username %><br>
+    <%= current_user.email %>
+  <% end %>
 </div>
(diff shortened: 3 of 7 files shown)


app/views/pages/index.html.erb
```diff
@@ -1,4 +1,21 @@
...
+    <h2>Logged-in User:</h2>
+    <hr>
+    <%= current_user.username %><br>
+    <%= current_user.email %>
+    <%= link_to 'log out', session_path, method: :delete %>
+  <% else %>
+    <h2>Welcome, stranger!</h2><br>
+      <div class="row">
+        <div class="col col-lg-1 col-lg-offset-5">
+          <%= link_to 'Log in', login_path %>
```
gregdaynes: pages#index fails to load, make sure you're using the right resource_path

app/views/pages/index.html.erb
```diff
@@ -1,4 +1,21 @@
...
+        </div>
+        <div class="col col-lg-1">
+          <%= link_to 'Sign up', signup_path %>
```
gregdaynes: pages#index fails to load without the correct signup resource
use `bin/rails routes` to view them

app/views/sessions/new.html.erb
```diff
@@ -1,2 +1,23 @@
...
+      <hr>
+      <div id="login_form">
+        <%= form_for :session, url: new_session_path  do |f| %>
+          <div class="form-group row">
+            <%= f.label :email, class: 'col-sm-4' %>
+            <%= f.text_field :email, class: 'col-sm-8' %>
+          </div>
+          <div class="form-group row">
+            <%= f.label :password, class: 'col-sm-4' %>
+            <%= f.text_field :password, class: 'col-sm-8' %>
```
gregdaynes: when entering my password, it is not masked

app/views/pages/index.html.erb
ghost: Hey, Greg, thanks for the feedback! Should be fixed by now.

app/helpers/sessions_helper.rb
```diff
@@ -0,0 +1,2 @@
+module SessionsHelper
```
qr8r, line 1: This commit adds an empty module to the project. It is not very useful.

Your next commit adds a login helper, which is very useful. I think this commit should be squashed with the one after it.

Remember: Commits need to be **functional**

app/helpers/sessions_helper.rb
```diff
@@ -2,4 +2,8 @@ module SessionsHelper
   def log_in(user)
     session[:user_id] = user.id
   end
+
+  def current_user
+    @current_user ||= User.find_by(id: session[:user_id])
```
qr8r: Nice use of memoization :)

app/helpers/sessions_helper.rb
```diff
@@ -3,6 +3,10 @@ def log_in(user)
     session[:user_id] = user.id
   end
 
+  def logged_in?
+    !current_user.nil?
```
qr8r: Double negative .... makes it confusing to read. 

In ruby, all values except for nil and false are considered 'truthy' values. You can just return current_user from this method.

Another 'fancy' way of doing this would be to alias the `current_user` method to `:logged_in?`

app/views/pages/index.html.erb
```diff
@@ -1,4 +1,10 @@
 <div class="container text-center">
   <h1>Pages#index</h1>
   <p>Find me in app/views/pages/index.html.erb</p>
+  <% if logged_in? %>
```
qr8r: Sort of covered in the comment above. This method is really just the same as `current_user`  its slightly more semantic to be able to use `logged_in?` but not by a lot.

Consider if you need the `logged_in?` method at all.

app/controllers/sessions_controller.rb
```diff
@@ -1,7 +1,15 @@
 class SessionsController < ApplicationController
   def new; end
 
-  def create; end
+  def create
+    user = User.find_by(email: params[:session][:email].downcase)
```
qr8r: If you are down casing here, make sure that you are also downcasing the email when it gets stored in the database. 

It may be something you care about, but this could also be a requirement that you decide is not worth implementing. Does it matter if a user tries to use upper cases in their email address?

app/views/sessions/new.html.erb
```diff
@@ -1,2 +1,23 @@
...
-<p>Find me in app/views/sessions/new.html.erb</p>
+<div class="container">
+  <div class="row">
+    <div class="col col-lg-4 offset-lg-4 text-center">
+      <h1>Log in</h1>
+      <hr>
+      <div id="login_form">
+        <%= form_for :session, url: new_session_path  do |f| %>
+          <div class="form-group row">
+            <%= f.label :email, class: 'col-sm-4' %>
```
qr8r: Not sure this is how the bootstrap grid is meant to be used. My understanding is it should go on container elements. 

It looks to me like you want a style to change the width of this element

app/views/sessions/new.html.erb
```diff
@@ -11,7 +11,7 @@
           </div>
           <div class="form-group row">
             <%= f.label :password, class: 'col-sm-4' %>
-            <%= f.text_field :password, class: 'col-sm-8' %>
+            <%= f.password_field :password, class: 'col-sm-8' %>
```
qr8r: Re: Greg's comment. This bit should be squashed into the commit where the form fields are introduced.

Try as much as possible not to fix commits with more commits within the same PR. Before it is merged you have full control of changing every commit within the PR, so every commit should be as pristine as you can make it. 

After a PR is merged into master, we mostly have no choice but to fix problems with additional commits. So its best to fix any bad commit history before things are merged when you have control over it.

qr8r (changes requested):

app/helpers/sessions_helper.rb
qr8r: Fixed

app/models/user.rb
```diff
@@ -4,5 +4,11 @@ class User < ApplicationRecord
   has_many :debts,   class_name: 'Debt', foreign_key: :debtor_id
   has_many :credits, class_name: 'Debt', foreign_key: :creditor_id
 
+  before_save :downcase_fields
+
+  def downcase_fields
```
qr8r: Why do we care that the email address is downcased?

app/views/sessions/new.html.erb
qr8r: This commecnt has still not addressed been addressed

app/views/sessions/new.html.erb
```diff
@@ -1,2 +1,37 @@
-<h1>Sessions#new</h1>
-<p>Find me in app/views/sessions/new.html.erb</p>
+<style>
```
qr8r: This is fine for now.

Css and javascript should be written in `app/assets`

CSS is applied based on an order of priority: 
- Inline stles
- In page styles
- external stylesheets

This is an example of an in page style. It will take precedence over any external styles making your page harder to maintain. All styles should be written as external stylesheets

qr8r (approved):
Approved with comments to be fixed up later

PR #23, opened 2017-02-16
Add functionality to edit user profiles

ghost:
Adds functionality to edit logged in user's profile.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -7,6 +7,10 @@ class UsersController < ApplicationController
     @user = User.new
   end
 
+  def edit
+    @user = User.find(params[:id])
+  end
+
   def show
     @user = User.find(params[:id])
   end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,4 +1,6 @@
 class UsersController < ApplicationController
+  before_action :verify_editor, only: [:edit, :update]
+
   def index
     @users = User.all
   end
@@ -39,5 +41,9 @@ class UsersController < ApplicationController
 
   def user_params
     params.require(:user).permit(:username, :email, :password, :avatar)
+  end
+
+  def verify_editor
+    redirect_to new_session_path unless current_user == User.find(params[:id])
   end
 end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -24,6 +24,17 @@ class UsersController < ApplicationController
     end
   end
 
+  def update
+    @user = User.find(params[:id])
+    if @user.update(user_params)
+      flash[:success] = 'Your information was successfully updated.'
+      redirect_to @user
+    else
+      flash[:danger] = 'Your information could not be updated.'
+      render 'edit'
+    end
+  end
+
   private
 
   def user_params
--- a/app/views/pages/index.html.erb
+++ b/app/views/pages/index.html.erb
@@ -6,16 +6,17 @@
     <hr>
     <%= current_user.username %><br>
     <%= current_user.email %>
-    <%= link_to 'log out', session_path, method: :delete %>
+    <%= link_to 'Log Out', session_path, method: :delete %>
+    <%= link_to 'Edit Profile', edit_user_path(current_user) %>
   <% else %>
-    <h2>Welcome, stranger!</h2><br>
-      <div class="row">
-        <div class="col col-lg-1 offset-lg-5">
-          <%= link_to 'Log in', new_session_path %>
-        </div>
-        <div class="col col-lg-1">
-          <%= link_to 'Sign up', new_user_path %>
-        </div>
+    <h2>Welcome, stranger!</h2>
+    <div class="row">
+      <div class="col col-lg-1 offset-lg-5">
+        <%= link_to 'Log in', new_session_path %>
       </div>
+      <div class="col col-lg-1">
+        <%= link_to 'Sign up', new_user_path %>
+      </div>
+    </div>
   <% end %>
 </div>
--- a/app/views/users/_user_form.html.erb
+++ b/app/views/users/_user_form.html.erb
@@ -1,0 +1,25 @@
+<%= form_for @user do |f| %>
+  <div class="form-group row">
+    <%= f.label :username, class: 'label sess-label' %>
+    <%= f.text_field :username, class: 'input-field sess-input-field' %>
+  </div>
+
+  <div class="form-group row">
+    <%= f.label :email, class: 'label sess-label' %>
+    <%= f.text_field :email, class: 'input-field sess-input-field' %>
+  </div>
+
+  <div class="form-group row">
+    <%= f.label :password, class: 'label sess-label' %>
+    <%= f.password_field :password, class: 'input-field sess-input-field' %>
+  </div>
+
+  <div class="form-group row">
+    <%= f.label :password_confirmation, class: 'label sess-label' %>
+    <%= f.password_field :password_confirmation, class: 'input-field sess-input-field' %>
+  </div>
+
+  <div class="form-group row">
+    <%= f.submit yield(:button_text), class: 'sess-button btn btn-primary' %>
+  </div>
+<% end %>
--- a/app/views/users/edit.html.erb
+++ b/app/views/users/edit.html.erb
@@ -1,0 +1,10 @@
+<% provide(:button_text, 'Save changes') %>
+<div class="container">
+  <div class="row text-center">
+    <div class="offset-lg-4">
+      <h2>Editing user <%= @user.username %></h2>
+      <hr>
+      <%= render 'user_form' %>
+    </div>
+  </div>
+</div>
--- a/app/views/users/new.html.erb
+++ b/app/views/users/new.html.erb
@@ -1,23 +1,10 @@
-<%= form_for @user do |f| %>
-  <p>
-    <%= f.label :username %>
-    <%= f.text_field :username %>
-  </p>
-
-  <p>
-    <%= f.label :email %>
-    <%= f.text_field :email %>
-  </p>
-
-  <p>
-    <%= f.label :password %>
-    <%= f.password_field :password %>
-  </p>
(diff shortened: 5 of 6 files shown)


app/views/pages/index.html.erb
```diff
@@ -1,4 +1,22 @@
...
   <p>Find me in app/views/pages/index.html.erb</p>
+  <% if logged_in? %>
+    <h2>Logged-in User:</h2>
+    <hr>
+    <%= current_user.username %><br>
+    <%= current_user.email %>
+    <%= link_to 'Log Out', session_path, method: :delete %>
+    <%= link_to 'Edit Profile', edit_user_path(current_user) %>
+  <% else %>
+    <h2>Welcome, stranger!</h2><br>
```
gregdaynes: Header elements in HTML are treated as blocks (clear on both sides). You can skip the < br > element in this case. Removing it will also make styling with CSS easier later on.

app/views/pages/index.html.erb
```diff
@@ -1,4 +1,22 @@
 <div class="container text-center">
   <h1>Pages#index</h1>
   <p>Find me in app/views/pages/index.html.erb</p>
+  <% if logged_in? %>
```
gregdaynes: undefined method on a new deploy

app/views/pages/index.html.erb
```diff
@@ -1,4 +1,22 @@
...
+    <h2>Logged-in User:</h2>
+    <hr>
```
gregdaynes: Decorator elements can be replaced by css.

ghost:
Won't function until the previous pull request is merged.

app/controllers/users_controller.rb
```diff
@@ -20,9 +26,22 @@ def create
     end
   end
 
+  def update
+    @user = User.find_by(params[:id])
+    if @user.update(user_params)
+      redirect_to @user
+    else
+      render 'edit'
```
jacquesporveau: Might be nice to add a flash or something here so that the user has an idea of why their update didn't work.

Also noticing you are using a `.find_by` call here. I think the difference between `.find` and `.find_by` is that `.find_by` doesn't error out if it doesn't find the object so there could be an instance where `@user` is nil on line 31 and calling `.update` would return a NoMethodError anyways. I would like to see consistency between using the two different types of find calls unless there is something I am missing and there is a specific reason you are using `.find_by` in this circumstance.

app/views/users/_user_form.html.erb
```diff
@@ -0,0 +1,23 @@
...
+  <p>
+    <%= f.label :email %>
+    <%= f.text_field :email %>
+  </p>
+
+  <p>
+    <%= f.label :password %>
+    <%= f.password_field :password %>
+  </p>
+  <p>
```
jacquesporveau: Inconsistent spacing. See line 11.

app/views/users/_user_form.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    <%= f.label :password_confirmation %>
+    <%= f.password_field :password_confirmation %>
+  </p>
```
jacquesporveau: Inconsistent spacing.

app/views/users/_user_form.html.erb
```diff
@@ -0,0 +1,23 @@
...
+  <p>
+  <%= f.submit yield(:button_text), class: 'btn btn-primary' %>
```
jacquesporveau: Indentation error.

app/views/users/edit.html.erb
```diff
@@ -0,0 +1,10 @@
+<% provide(:button_text, 'Save changes') %>
+<div class="container">
+  <div class="row text-center">
+    <div class="offset-lg-4">
+      <h2>Editing user <%= @user.username %> </h2>
```
jacquesporveau: I personally don't like the space between the closing erb tag and closing h2 tag. Not a huge thing but it just doesn't need to be there imo.

app/views/pages/index.html.erb
```diff
@@ -1,4 +1,23 @@
 <div class="container text-center">
   <h1>Pages#index</h1>
   <p>Find me in app/views/pages/index.html.erb</p>
+  <% if current_user %>
+    <h2>Logged-in User:</h2>
+    <hr>
+    <%= current_user.username %><br>
+    <%= current_user.email %>
+    <h2>Welcome, stranger!</h2><br>
```
jacquesporveau: If the current user is logged in we are saying welcome, stranger? This is also in your else block. Something doesn't add up.

app/views/pages/index.html.erb
```diff
@@ -1,4 +1,23 @@
 <div class="container text-center">
   <h1>Pages#index</h1>
   <p>Find me in app/views/pages/index.html.erb</p>
+  <% if current_user %>
```
jacquesporveau: Does this work? I believe in order for a variable to be in scope for a view it needs to be an instance variable.

jacquesporveau (dismissed):

app/views/pages/index.html.erb
```diff
@@ -1,4 +1,23 @@
...
+    <hr>
+    <%= current_user.username %><br>
+    <%= current_user.email %>
+    <h2>Welcome, stranger!</h2><br>
+    <%= link_to 'Log Out', session_path, method: :delete %>
+    <%= link_to 'Edit Profile', edit_user_path(current_user) %>
+  <% else %>
+    <h2>Welcome, stranger!</h2>
+      <div class="row">
+        <div class="col col-lg-1 offset-lg-5">
```
jacquesporveau: Noticing inconsistencies between double quotations and single. I think you are supposed to use single quotations everywhere that isn't using string interpolation.

qr8r:
This PR needs a description

app/views/pages/index.html.erb
ghost: Hey, thank you for all the feedback! I'll do my best to fix it up.
Yeah, I am using single quotes for Ruby stuff and double quotes for JavaScript/HTML, is that considered a bad style or? Brandon said it's okay as long as it's consistent between the two, but I have no idea, so thanks in advance.

app/views/pages/index.html.erb
ghost: This will work when the pull request gets merged with the previous one, yeah. As it is it will give an error.

app/views/users/_user_form.html.erb
```diff
@@ -0,0 +1,41 @@
+<style>
```
qr8r: Same comment as before 

Styles should always be defined as external stylesheets in `app/assets`

app/views/users/new.html.erb
```diff
@@ -1,23 +1,10 @@
...
-    <%= f.submit 'Sign up' %>
-  </p>
-<% end %> 
+<% provide(:button_text, 'Sign Up') %>
+<div class="container">
+  <div class="row text-center">
+    <div class="offset-lg-4">
+      <h2>Creating New User</h2>
+      <hr>
+      <%= render 'users/user_form', user: @user %>
```
qr8r: This can just be `render :user_form`.

Since the view you are rendering (new) lives in the users directory, Rails will look in that directory for the partial you are referencing. You only need to define the full path to a partial in 2 circumstances: 

- you are referencing a partial from a partial (nested partials) 
- You are referencing a partial that doesnt live in the same view path as the view you are currently renderign 
  - ie. im rendering /bets/show and I want to render /users/_user_information

app/views/pages/index.html.erb
qr8r: @jacquesporveau @YarekTenko Yarek is correct. current_user is a method defined in his sessions_helper and as such it will included in all view contexts

app/views/pages/index.html.erb
qr8r: Consistency is key. At the end of the day quotations doesnt really matter so long as you are consistent.

app/views/users/edit.html.erb
```diff
@@ -0,0 +1,10 @@
+<% provide(:button_text, 'Save changes') %>
+<div class="container">
+  <div class="row text-center">
+    <div class="offset-lg-4">
+      <h2>Editing user <%= @user.username %></h2>
+      <hr>
+      <%= render 'users/user_form', user: @user %>
```
qr8r: can just be `render :user_form`

qr8r (changes requested):
small changes, then this is good to go

app/views/pages/index.html.erb
qr8r: as a clarification, always double quotes in html, and single quotes in ruby with double for interpolation is consistent. just for different contexts. The erb tags make this a little confusing

qr8r:
This PR is dependent on #22

ghost:
Changed :user_form to 'user_form'.

app/views/pages/index.html.erb
```diff
@@ -6,9 +6,10 @@
     <hr>
     <%= current_user.username %><br>
     <%= current_user.email %>
-    <%= link_to 'log out', session_path, method: :delete %>
+    <%= link_to 'Log Out', session_path, method: :delete %>
+    <%= link_to 'Edit Profile', edit_user_path(current_user) %>
   <% else %>
-    <h2>Welcome, stranger!</h2><br>
+    <h2>Welcome, stranger!</h2>
       <div class="row">
```
gregdaynes: Minor indentation note:
In html, if your code following an element that is self closing, or has a closing tag. 
it is clearer to indicat a sibling (not child) element by having it at the same indentation.

```
<h2>Welcome, Stranger!</h2>
<div> <!-- sibling of h2 -->
...
</div>
```

qr8r (approved):

qr8r (changes requested):
Broken pending #27

ghost:
#27 Has been merged, seems like all the other issues have been addressed too.

qr8r (approved):

qr8r:
This commit has a merge commit that shoudnt be there. Please remove it before merging

PR #24, opened 2017-02-22
Add spec for BetOption model

ghost:
Defined winning_option_total method for the BetOption model and provided an rspec for testing it.

### DIFF
--- a/app/models/bet_option.rb
+++ b/app/models/bet_option.rb
@@ -4,4 +4,8 @@ class BetOption < ApplicationRecord
 
   scope :winners, -> { where(winner: true) }
   scope :losers, -> { where(winner: false) }
+
+  def option_total
+    user_bets.pluck(:amount_bet).sum
+  end
 end
--- a/spec/models/bet_option_spec.rb
+++ b/spec/models/bet_option_spec.rb
@@ -1,0 +1,62 @@
+require 'rails_helper'
+
+RSpec.describe BetOption, type: :model do
+  let!(:bob) do
+    User.create(username: 'Bob', email: 'bob@mail.com', password: '123')
+  end
+  let!(:peter) do
+    User.create(username: 'Peter', email: 'peter@mail.com', password: 'pete1')
+  end
+  let!(:jack) do
+    User.create(username: 'Jack', email: 'jack@mail.com', password: 'notjack')
+  end
+
+  let(:bet1) do
+    Bet.create(bet: 'Superbowl: Atlanta vs New England, place your bets!')
+  end
+
+  let!(:users_place_bets) do
+    UserBet.create(user:       bob,
+                   bet:        bet1,
+                   bet_option: option1,
+                   amount_bet: 100)
+
+    UserBet.create(user:       peter,
+                   bet:        bet1,
+                   bet_option: option1,
+                   amount_bet: 50)
+
+    UserBet.create(user:       jack,
+                   bet:        bet1,
+                   bet_option: option2,
+                   amount_bet: 300)
+  end
+
+  describe('#option_total') do
+    subject { winning_option.option_total }
+
+    context 'Atlanta wins' do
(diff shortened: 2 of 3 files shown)


qr8r:
^^  Needs description

app/models/bet_option.rb
```diff
@@ -2,6 +2,12 @@ class BetOption < ApplicationRecord
   has_many :users, through: :user_bets
   has_many :user_bets
 
+  def winner_total
```
qr8r: I dont think this should be winner_total. Not all bets are winners,

Also, instance methods should be placed below class methods (scopes)

spec/models/bet_option_spec.rb
```diff
@@ -0,0 +1,58 @@
+require 'rails_helper'
+
+RSpec.describe BetOption, type: :model do
```
qr8r, line 3: Adding a method, and the spec for that method in a single commit is preferred. 

We track something called code coverage on projects, which is the percentage of your code that gets run during your test suite. We strive to keep all of our Repos at 100% code coverage.

If you add a method in one commit without the associated specs for it, then the code coverage drops for that commit specifically

spec/models/bet_option_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+                   bet:        bet1,
+                   bet_option: option2,
+                   amount_bet: 300)
+  end
+
+  describe('#winner_total') do
+    subject { winning_option.winner_total }
+
+    context 'Atlanta wins' do
+      let!(:atlanta_won)   { option1.update(winner: true) }
```
qr8r: Might be better to define the option here instead of defining it above and updating it here

qr8r (commented):
Looking pretty good! Again just a few small review comments

app/models/bet_option.rb
```diff
@@ -4,4 +4,10 @@ class BetOption < ApplicationRecord
 
   scope :winners, -> { where(winner: true) }
   scope :losers, -> { where(winner: false) }
+
+  def winning_option_total
```
qr8r: I think this method should just be calculating the total bet. Not all bets are winners

qr8r (changes requested):
I think the PR is dependent on #16 ( or possibly the other way around) 1 PR breaks the other... we will probably want to merge one of the 2 first, and then fixup the other one

ghost:
```
   def amount_owed(lost_bet, won_bet)
-    lost_bet.amount_bet * (won_bet.amount_bet / winner_total).round(2)
+    won_option = options.find_by(winner: true).option_total
+    lost_bet.amount_bet * (won_bet.amount_bet / won_option).round(2)
   end
```
Fixes the dependency with #16 (it has already been merged).

jacquesporveau:
Just a heads up you have a merge commit as your last commit. You can avoid this by running `$ git rebase --onto origin master`

qr8r:
Git rebase --onto is da :bomb: 

https://git-scm.com/docs/git-rebase

Can be a little tricky at first

app/models/bet_option.rb
```diff
@@ -4,4 +4,8 @@ class BetOption < ApplicationRecord
 
   scope :winners, -> { where(winner: true) }
   scope :losers, -> { where(winner: false) }
+
+  def option_total
+    user_bets.map(&:amount_bet).sum
```
qr8r: Doesnt need to change but just pointing out that this calculation is happening in the application layer.

If you wanted it to happen at the database layer your could use `user_bets.pluck(:amount_bet).sum` I super recommend looking up the `pluck` docs, it can be a super useful tool to use.

Performing the calc in either the app layer or database layer has advantages and disadvantages. I dont really know what is most important here so there is no need to change it. (but maybe try and think what the advantages and disadvantages could be)

qr8r (approved):

PR #27, opened 2017-02-24
Fix Tether

ghost:
Produced "Not Found" error before.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -11,8 +11,7 @@
  * It is generally better to create a new file per style scope.
  *
 
- *= require tether-theme-basic
- *= require tether-theme-arrows
+ *= require tether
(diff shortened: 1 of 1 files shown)


ghost:
Merging "Add Tether Gem" into the master branch resulted into an "file not found error". This PR fixes that issue.

gregdaynes:
What is your use-case that you need to use Tether? I've never used it before, and am curious about the necessity of using a JS library to align things after DOM generation, instead of using proper markup + css.

qr8r:
@gregdaynes Tether is a dependency of bootstrap 4 
http://stackoverflow.com/questions/34567939/how-to-fix-the-error-error-bootstrap-tooltips-require-tether-http-github-h 

we might want to consider adding our JS dependencies as script tags, instead of as ruby gems

qr8r (approved):

PR #28, opened 2017-02-24
Add has_many/belongs_to relationship between Bet and BetOption

ghost:
Now every BetOption knows which bet it belongs to.

### DIFF
--- a/app/models/bet_option.rb
+++ b/app/models/bet_option.rb
@@ -1,6 +1,7 @@
 class BetOption < ApplicationRecord
   has_many :users, through: :user_bets
   has_many :user_bets
+  belongs_to :bet
 
   scope :winners, -> { where(winner: true) }
   scope :losers, -> { where(winner: false) }
--- a/db/migrate/20170220212242_add_bet_id_to_options.rb
+++ b/db/migrate/20170220212242_add_bet_id_to_options.rb
@@ -1,0 +1,7 @@
+class AddBetIdToOptions < ActiveRecord::Migration[5.0]
+  def change
+    add_column :bet_options, :bet_id, :integer
+    add_foreign_key :bet_options, :bets
+    add_index :bet_options, :bet_id
+  end
+end
--- a/app/models/bet.rb
+++ b/app/models/bet.rb
@@ -1,7 +1,7 @@
 class Bet < ApplicationRecord
   has_many :users, through: :user_bets
   has_many :user_bets
-  has_many :options, through: :user_bets, source: :bet_option
+  has_many :options, class_name: 'BetOption'
 
   def resolve(winning_option:)
     winning_option.update!(winner: true)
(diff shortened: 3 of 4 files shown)


gregdaynes (approved):

db/migrate/20170220212242_add_bet_id_to_options.rb
```diff
@@ -0,0 +1,6 @@
+class AddBetIdToOptions < ActiveRecord::Migration[5.0]
+  def change
+    add_column :bet_options, :bet_id, :integer
```
qr8r, line 3: Missing index

app/models/bet_option.rb
```diff
@@ -1,6 +1,7 @@
 class BetOption < ApplicationRecord
   has_many :users, through: :user_bets
   has_many :user_bets
+  belongs_to :bet
```
qr8r, line 4: I am ok with this, but if you are changing the model architechture, you may want to look at the rest of the relationships involved as well.

Currently: Bet has many options through user bets, but bet option belongs to a bet directly.

qr8r (changes requested):

YarekTenko:
Forgot to add foreign key to BetOption, at this stage it will produce: "can't write unknown attribute user_bet_id" error.

db/migrate/20170220212242_add_bet_id_to_options.rb
```diff
@@ -0,0 +1,7 @@
...
+    add_foreign_key :bet_options, :bets
+    add_index :bet_options, :bet_id
```
isaacfreeman, line 5: 👍

YarekTenko:
Problem: changing `has_many :options, through: :user_bets, source: :bet_option` to `has_many :options, class_name: 'BetOption'` without adding appropriate id column (I think).

app/models/user_bet.rb
```diff
@@ -2,6 +2,7 @@ class UserBet < ApplicationRecord
   belongs_to :user
   belongs_to :bet
   belongs_to :bet_option
+  has_one    :bet_option
```
qr8r: This is wrong

db/migrate/20170228184631_add_foreign_key_to_bet_option.rb
```diff
@@ -0,0 +1,7 @@
+class AddForeignKeyToBetOption < ActiveRecord::Migration[5.0]
+  def change
+    add_column :bet_options, :user_bet_id, :integer
```
qr8r: Why do bet options belong to user bets?

qr8r (changes requested):
BetOptions should not have a user id

isaacfreeman (approved):

qr8r (approved):
Requires Rebase

PR #30, opened 2017-02-24
Create Bets

ghost:
Dependent on #28
Allows logged in users to create bets by specifying their text and expiry date.
Created bets can be viewed separately.
All bets will be available through /bets page, where user will be able to specify search criteria to filter through the database.
Ongoing (active bets with status set to 1) will be displayed on the main page

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -14,5 +14,6 @@
 //= require tether
 //= require bootstrap-sprockets
 //= require jquery_ujs
+//= require lib/date_picker
 //= require turbolinks
 //= require_tree .
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -20,3 +20,4 @@
 @import "bootstrap";
 @import "font-awesome";
 @import "session-access";
+@import "user-bar";
--- a/app/controllers/bets_controller.rb
+++ b/app/controllers/bets_controller.rb
@@ -1,4 +1,8 @@
 class BetsController < ApplicationController
+  def show
+    @bet = Bet.find(params[:id])
+  end
+
   def new
     @bet = Bet.new
   end
--- a/app/controllers/bets_controller.rb
+++ b/app/controllers/bets_controller.rb
@@ -1,4 +1,8 @@
 class BetsController < ApplicationController
+  def index
+    @bets = Bet.all
+  end
+
   def show
     @bet = Bet.find(params[:id])
   end
--- a/app/controllers/bets_controller.rb
+++ b/app/controllers/bets_controller.rb
@@ -1,0 +1,5 @@
+class BetsController < ApplicationController
+  def new
+    @bet = Bet.new
+  end
+end
--- a/app/controllers/bets_controller.rb
+++ b/app/controllers/bets_controller.rb
@@ -10,4 +10,21 @@ class BetsController < ApplicationController
   def new
     @bet = Bet.new
   end
+
+  def create
+    @bet = Bet.new(bet_params)
+    if @bet.save
+      flash[:success] = 'Bet has been created'
+      redirect_to @bet
+    else
+      flash[:danger] = 'Could not create bet'
+      render 'new'
+    end
+  end
+
+  private
+
+  def bet_params
+    params.require(:bet).permit(:bet, :status, :expires_at, :creator_id)
+  end
 end
--- a/app/views/bets/new.html.erb
+++ b/app/views/bets/new.html.erb
@@ -1,0 +1,27 @@
+<div class="container">
+  <hr>
+  <% if current_user %>
+    <h2> Creating a new bet </h2>
+    <hr>
+    <%= form_for @bet do |f| %>
+      <%= f.hidden_field :creator_id, value: current_user.id %>
+
+      <div class="form-group">
+        <%= f.label 'Enter bet text:' %>
+        <%= f.text_field :bet, class: 'form-control' %>
+      </div>
+
+      <div class="form-group">
+        <%= f.label 'Enter bet expiration date (placing your bets is not allowed after this date:' %>
+        <%= f.text_field :expires_at, class: 'form-control jquery-datepicker' %>
+      </div>
+
+      <div class="form-group">
+        <%= f.submit 'Create bet', class: 'btn btn-primary' %>
+      </div>
+    <% end %>
+
+  <% else %>
+    <h2>You are not logged in </h2>
+  <% end %>
+</div>
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -6,6 +6,11 @@
 
     <%= stylesheet_link_tag    'application', media: 'all', 'data-turbolinks-track': 'reload' %>
     <%= javascript_include_tag 'application', 'data-turbolinks-track': 'reload' %>
+
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery-datetimepicker/2.5.4/build/jquery.datetimepicker.full.min.js"></script>
+
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/jquery-datetimepicker/2.5.4/build/jquery.datetimepicker.min.css" />
+    <link href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet">
   </head>
 
   <body>
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -13,6 +13,7 @@
       <% flash.each do |key, value| %>
         <div class="alert alert-<%= key %>"><%= value %></div>
       <% end %>
+      <%= render 'shared/user_bar' %>
       <%= yield %>
     </div>
   </body>
--- a/app/views/pages/index.html.erb
+++ b/app/views/pages/index.html.erb
@@ -1,15 +1,14 @@
-<div class="container text-center">
-  <h1>Pages#index</h1>
-  <p>Find me in app/views/pages/index.html.erb</p>
+<div class="container">
   <% if current_user %>
-    <h2>Logged-in User:</h2>
     <hr>
-    <%= current_user.username %><br>
-    <%= current_user.email %>
-    <%= link_to 'Log Out', session_path, method: :delete %>
-    <%= link_to 'Edit Profile', edit_user_path(current_user) %>
-  <% else %>
-    <h2>Welcome, stranger!</h2>
+  <% end %>
+
+  <% unless current_user %>
+    <div class="row text-center">
+      <div class="col col-lg-4 offset-lg-4">
+        <h2>Welcome, stranger!</h2><br>
+      </div>
+    </div>
     <div class="row">
       <div class="col col-lg-1 offset-lg-5">
         <%= link_to 'Log in', new_session_path %>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -8,5 +8,6 @@ Rails.application.routes.draw do
 
   get '/signup', to: 'users#new', as: :new_user_path
 
+  resources :bets
   resources :users
 end
--- a/app/assets/javascripts/lib/date_picker.js
+++ b/app/assets/javascripts/lib/date_picker.js
@@ -1,0 +1,3 @@
+$(document).on("turbolinks:load", function(){
+  $(".jquery-datepicker").datetimepicker();
+});
--- a/app/assets/stylesheets/user-bar.scss
+++ b/app/assets/stylesheets/user-bar.scss
@@ -1,0 +1,5 @@
+.user-bar{
+  img{
+    width: 50px;
+  }
+}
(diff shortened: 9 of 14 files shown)


app/views/bets/new.html.erb
```diff
@@ -0,0 +1,30 @@
...
+      </div>
+
+      <div class="form-group">
+        <button type="submit" class="btn btn-primary">Create bet</button>
+      </div>
+    </form>
+
+  <% else %>
+    <h2>You are not logged in </h2>
+<% end %>
```
jacquesporveau: Identation on the `end`

Also: curious why we aren't using a Rails form_for tag?

app/views/layouts/application.html.erb
```diff
@@ -9,6 +9,10 @@
   </head>
 
   <body>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery-datetimepicker/2.5.4/build/jquery.datetimepicker.full.min.js"></script>
+
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/jquery-datetimepicker/2.5.4/build/jquery.datetimepicker.min.css" />
+    <link href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet">
```
jacquesporveau: Looks like Rails is loading the other js and css in the head tag. Putting these in the body seems a bit odd.

app/views/pages/index.html.erb
```diff
@@ -1,14 +1,55 @@
...
   <% if current_user %>
-    <h2>Logged-in User:</h2>
+    <div class="row">
+      <div class="col col-lg-4">
+        <%= link_to 'Create Bet', new_bet_path %>
+      </div>
+      <div class="col col-lg-4">
+      </div>
+      <div class="col col-lg-3">
+      </div>
```
jacquesporveau: Up to you for style preferences but I think we could single line these last couple lines
`<div class="col col-lg-3></div>` etc.

app/views/pages/index.html.erb
```diff
@@ -1,14 +1,55 @@
...
+      <div class="col col-lg-1">
+        <div class="dropdown">
+            <img class="" type="button" data-toggle="dropdown" src="http://fuuse.net/wp-content/uploads/2016/02/avatar-placeholder.png" style="width: 100%;">
```
jacquesporveau: 4 space indent and not sure we aren't using the Rails image_tag?
http://apidock.com/rails/ActionView/Helpers/AssetTagHelper/image_tag
Also noticing you have a class attr with no data. Not sure if needed?

app/views/pages/index.html.erb
```diff
@@ -1,14 +1,55 @@
...
+        <h2>Ongoing bets</h2>
+      </div>
+    </div>
+    <div class="row">
+      <table class="table table-hover">
+        <thead class="thead-default">
+          <tr>
+            <th class="col-md-2">
+              Created By
+            </th>
```
jacquesporveau: You can one line these all if you want. Up to you but I think it looks better.

app/views/pages/index.html.erb
```diff
@@ -1,14 +1,55 @@
...
+      </table>
+    </div>
   <% else %>
-    <h2>Welcome, stranger!</h2><br>
+    <div class="row text-center">
+      <div class="col col-lg-4 offset-lg-4">
+        <h2>Welcome, stranger!</h2><br>
+      </div>
+    </div>
       <div class="row">
```
jacquesporveau: Not part of the PR but your identation is off here too.

jacquesporveau (commented):
Looking good! Left some general feedback. Most of it is optional.

app/views/layouts/application.html.erb
ghost: Thank you very much for all the feedback, I'll get to fixing it!
Yeah, I actually tried putting it in the head tag but it wouldn't load for some reason. Maybe I did something wrong, I'll try again.

app/views/bets/new.html.erb
ghost: Good catch, thanks! 
Just feels like doing it manually gives me a bit more control over the style, but I might be mistaken here. Is it considered a better idea to use form_for in situations like that?

jacquesporveau: It's considered to be best practices. It's kind of like buying a state of the art chainsaw and then using a handsaw instead.
I think some of the philosophy behind Rails is "lets make things easy". By using the form for tag you are using the Rails magic.

app/views/bets/new.html.erb
```diff
@@ -0,0 +1,30 @@
+<div class="container">
+  <% if current_user %>
+    <h2> Creating a new bet </h2>
+    <form action="<% create_bet_path %>" method="post" id="create-bet-form">
```
isaacfreeman: In situations like this where there's no instance to to use with `form_for`, Rails provides `form_tag` and corresponding methods like `text_field_tag`, `hidden_field_tag` etc.

app/views/pages/index.html.erb
```diff
@@ -1,21 +1,52 @@
...
+<div class="container">
   <% if current_user %>
-    <h2>Logged-in User:</h2>
+    <div class="row">
+      <div class="col col-lg-4">
+        <%= link_to 'Create Bet', new_bet_path %>
+      </div>
+      <div class="col col-lg-4"></div>
+      <div class="col col-lg-3"></div>
+      <div class="col col-lg-1">
```
isaacfreeman: Can you use a Bootstrap offset class here to replace the empty tags above?

app/views/pages/index.html.erb
```diff
@@ -1,21 +1,52 @@
...
+        <div class="dropdown">
+          <img type="button" data-toggle="dropdown" src="http://fuuse.net/wp-content/uploads/2016/02/avatar-placeholder.png" style="width: 100%;">
```
isaacfreeman: It looks like this image represents a page control. It's probably best to use a `<button>` here with the image specified in the CSS, and keep `` for images that are content rather than controls.

config/routes.rb
```diff
@@ -8,5 +8,7 @@
 
   get '/signup', to: 'users#new', as: :new_user_path
 
+  get  '/create_bet', to: 'bets#new', as: :new_bet
```
qr8r: You should be using the `resources` syntax in order to create RESTful routes. 

http://guides.rubyonrails.org/routing.html

app/controllers/bets_controller.rb
```diff
@@ -1,5 +1,22 @@
 class BetsController < ApplicationController
-  def new_bet
+  def new
```
qr8r: Should be squashed

app/views/pages/index.html.erb
```diff
@@ -1,21 +1,52 @@
...
+      </div>
+    </div>
     <hr>
-    <%= current_user.username %><br>
-    <%= current_user.email %>
-    <%= link_to 'log out', session_path, method: :delete %>
+
+    <div class="row">
+      <div class="col col-md-4">
+        <h2>Ongoing bets</h2>
```
qr8r: This content would be better suited for a bet/index page now

app/views/pages/index.html.erb
```diff
@@ -1,21 +1,52 @@
...
+      <div class="col col-lg-4">
+        <%= link_to 'Create Bet', new_bet_path %>
+      </div>
+      <div class="col col-lg-4"></div>
+      <div class="col col-lg-3"></div>
+      <div class="col col-lg-1">
+        <div class="dropdown">
+          <img type="button" data-toggle="dropdown" src="http://fuuse.net/wp-content/uploads/2016/02/avatar-placeholder.png" style="width: 100%;">
+          <ul class="dropdown-menu">
+            <li><%= link_to 'Log Out', session_path, method: :delete %></li>
```
qr8r: Since this is content that is likely shared across many pages, It should be extracted into a partial and rendered in the application layout

app/controllers/bets_controller.rb
```diff
@@ -1,5 +1,22 @@
...
   end
+
+  def create
+    @bet = Bet.new(bet_params)
+    if @bet.save
+      flash[:success] = 'Bet has been created'
+      redirect_to root_path
+    else
+      flash[:danger] = 'Could not create bet'
+      redirect_to create_bet_path
```
qr8r: Should render the create page so that the form fields stay populated with the values the user tried to use.

app/controllers/bets_controller.rb
```diff
@@ -1,5 +1,22 @@
...
-  def new_bet
+  def new
     @bet = Bet.new
   end
+
+  def create
+    @bet = Bet.new(bet_params)
+    if @bet.save
+      flash[:success] = 'Bet has been created'
+      redirect_to root_path
```
qr8r: Should redirect to the Bet#show page
Same with update when you get there. 

After you create something, as a user, dont you want to see the thing you created?

app/views/bets/new.html.erb
```diff
@@ -0,0 +1,30 @@
...
+      <div>
+        <input type="hidden" value="<%= form_authenticity_token %>" name="authenticity_token"/>
```
qr8r: You might want to stick to the rails form helpers. They provide a lot of begin the scenes help for you. 

Once example, is that if you try and edit an existing record. All of the records fields will be pre populated in the form fields
http://guides.rubyonrails.org/form_helpers.html

config/routes.rb
```diff
@@ -9,6 +9,7 @@
   get '/signup', to: 'users#new', as: :new_user_path
 
   get  '/create_bet', to: 'bets#new', as: :new_bet
+  post '/create_bet', to: 'bets#create', as: :create_bet
```
qr8r: use the rails `:resources` syntax. Explicitly defining routes with 'get' etc should be exceptional and used only when absolutely needed

qr8r (changes requested):
I would like this PR to include a Bet index page and a Bet show page.

They dont need to be fancy or styled. Just present and functional

isaacfreeman (approved):

app/views/pages/index.html.erb
ghost: My idea was to have all the current ongoing (active) bets (sorted by date) listed right on the front page and all the bets ever created on bet/index (with the ability to search/filter/etc.). Would it make sense to leave it here for now in this case?

YarekTenko:
Thanks for the feedback, everybody, hope I fixed most of the problems this time.

qr8r (changes requested):
This PR breaks the site

app/assets/stylesheets/application.scss
```diff
@@ -20,3 +20,4 @@
 @import "bootstrap";
 @import "font-awesome";
 @import "session-access";
+@import "user-bar";
```
qr8r, line 23: This commit needs a commit message

Adding a controller action and styling page elements should be two separate commits. 
Remember a commit is the smallest piece of coherent and functional code you can put together.

Adding a controller action and Adding a top navigation bar are unrelated and should not be in the same PR, let alone the same commit

app/views/pages/index.html.erb
```diff
@@ -1,15 +1,20 @@
...
-    <%= link_to 'Log Out', session_path, method: :delete %>
-    <%= link_to 'Edit Profile', edit_user_path(current_user) %>
-  <% else %>
-    <h2>Welcome, stranger!</h2>
+    <div class="row">
+      <div class="col col-md-4">
+        <h2>Ongoing bets</h2>
+      </div>
+    </div>
+    <%= render 'bets/display_bets' %>
```
qr8r: This is beyond the scope of the story. 

The story asks only for a page you can visit to create a new bet

config/routes.rb
```diff
@@ -8,5 +8,6 @@
 
   get '/signup', to: 'users#new', as: :new_user_path
 
+  resources :bets
```
qr8r, line 11: Are all bet routes available to you at this point?

app/controllers/sessions_controller.rb
```diff
@@ -5,7 +5,7 @@ def create
     user = User.find_by(email: params[:session][:email])
     if user && user.authenticate(params[:session][:password])
```
qr8r: ^^ Commit message is empty 

Why are you making this change?

app/assets/javascripts/application.js
```diff
@@ -14,5 +14,6 @@
 //= require tether
```
qr8r, line 14: This is a good sized commit

qr8r (changes requested):

app/views/layouts/application.html.erb
```diff
@@ -13,6 +13,7 @@
       <% flash.each do |key, value| %>
         <div class="alert alert-<%= key %>"><%= value %></div>
       <% end %>
+      <%= render partial: 'shared/user_bar' %>
```
qr8r: I believe you dont need to specify partial here.

from inside a view file rails automatically assumes it will be a partial.

<%= render shared/user_bar' %>

qr8r (approved):
Much better PR! 

One tiny comment, but its nto enough to ask for changes here

PR #31, opened 2017-02-24
Define bet_total for Bet

ghost:
Defined bet_total method for Bet and provided rspec for it.

### DIFF
--- a/app/models/bet.rb
+++ b/app/models/bet.rb
@@ -22,4 +22,8 @@ class Bet < ApplicationRecord
   def amount_owed(lost_bet, won_bet)
     lost_bet.amount_bet * (won_bet.amount_bet / winner_total).round(2)
   end
+
+  def bet_total
+    user_bets.pluck(:amount_bet).sum
+  end
 end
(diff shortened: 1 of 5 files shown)


gregdaynes (approved):
Looks good. One thing to look into with regards to testing is RSpec's one-line format.

app/models/bet.rb
```diff
@@ -22,4 +22,8 @@ def winner_total
   def amount_owed(lost_bet, won_bet)
     lost_bet.amount_bet * (won_bet.amount_bet / winner_total).round(2)
   end
+
+  def bet_total
+    user_bets.map(&:amount_bet).sum
```
qr8r: This is correct.

You could also do this with `user_bets.pluck(:amount_bet).sum` Which doesnt instantiate UserBet objects and performs the sum at the database level, instead of the application level

spec/models/bet_spec.rb
```diff
@@ -60,4 +60,13 @@
       end
     end
   end
+
+  describe('#bet_total') do
+    subject { bet1.bet_total }
+
+    let(:expected_bet_total) { 450 }
+    it 'returns the total amount of money placed on this bet' do
+      expect(subject).to eq expected_bet_total
```
qr8r: Re greg's comment: https://www.relishapp.com/rspec/rspec-core/docs/subject/one-liner-syntax

PR #32, opened 2017-02-27
Add validation for Create Bets

ghost:
Depends on #30 
-Validates presence of expires_at and creator_id fields for Bet.
-Validates the expiry date (can't be in the future) of expires_at column for Bet with custom ExpiryDateValidator.

### DIFF
--- a/app/controllers/bets_controller.rb
+++ b/app/controllers/bets_controller.rb
@@ -17,7 +17,7 @@ class BetsController < ApplicationController
       flash[:success] = 'Bet has been created'
       redirect_to @bet
     else
-      flash[:danger] = 'Could not create bet'
+      flash[:danger] = @bet.errors.full_messages.to_sentence
       render 'new'
     end
   end
--- a/app/models/bet.rb
+++ b/app/models/bet.rb
@@ -3,6 +3,9 @@ class Bet < ApplicationRecord
   has_many :user_bets
   has_many :options, class_name: 'BetOption'
 
+  validates :expires_at, presence: true
+  validates_with ExpiryDateValidator
+
   def resolve(winning_option:)
     winning_option.update!(winner: true)
     debt_relationships = user_bets.winners.to_a.product(user_bets.losers.to_a)
--- a/app/models/bet.rb
+++ b/app/models/bet.rb
@@ -4,6 +4,7 @@ class Bet < ApplicationRecord
   has_many :options, class_name: 'BetOption'
 
   validates :expires_at, presence: true
+  validates :creator_id, presence: true
   validates_with ExpiryDateValidator
 
   def resolve(winning_option:)
--- a/app/validators/expiry_date_validator.rb
+++ b/app/validators/expiry_date_validator.rb
@@ -1,0 +1,6 @@
+class ExpiryDateValidator < ActiveModel::Validator
+  def validate(record)
+    return unless record.expires_at && record.expires_at <= DateTime.current
+    record.errors[:expires_at] << 'date must be in the future'
+  end
+end
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -16,7 +16,12 @@
   <body>
     <div class="container">
       <% flash.each do |key, value| %>
-        <div class="alert alert-<%= key %>"><%= value %></div>
+        <div class="alert alert-<%= key %>">
+          <button type="button" class="close" data-dismiss="alert" aria-label="Close">
+            <i class="fa fa-times"></i>
+          </button>
+          <%= value %>
+        </div>
       <% end %>
       <%= render 'shared/user_bar' %>
       <%= yield %>
--- a/app/assets/javascripts/lib/date_picker.js
+++ b/app/assets/javascripts/lib/date_picker.js
@@ -1,3 +1,5 @@
 $(document).on("turbolinks:load", function(){
-  $(".jquery-datepicker").datetimepicker();
+  $(".jquery-datepicker").datetimepicker({
+    minDate: "0"
+  });
 });
--- a/db/migrate/20170303184553_add_presence_validation_to_creator_id.rb
+++ b/db/migrate/20170303184553_add_presence_validation_to_creator_id.rb
@@ -1,0 +1,5 @@
+class AddPresenceValidationToCreatorId < ActiveRecord::Migration[5.0]
+  def change
+    change_column_null :bets, :creator_id, false
+  end
+end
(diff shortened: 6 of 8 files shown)


jacquesporveau:
You have files showing up in your diff that were edited in other PR's. Might be an idea when you have these dependent PRs to have your PR set to merge into it's dependancy so that we can get a clear look at the diff else we will just be reviewing more code than you intended.

app/validators/expiry_date_validator.rb
```diff
@@ -0,0 +1,10 @@
+class ExpiryDateValidator < ActiveModel::Validator
+  def validate(record)
+    if !record.expires_at.nil? && record.expires_at <= DateTime.now
```
jacquesporveau: On this line you are checking if record.expires_at is present or not nil.
Since nil is falsy in ruby we can just check `if record.expires_at && record.expires_at <= DateTime.now`

app/views/layouts/application.html.erb
```diff
@@ -9,6 +9,20 @@
...
 
   <body>
-    <%= yield %>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery-datetimepicker/2.5.4/build/jquery.datetimepicker.full.min.js"></script>
+
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/jquery-datetimepicker/2.5.4/build/jquery.datetimepicker.min.css" />
+    <link href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet">
+    <div class="container">
+      <% flash.each do |key, value| %>
+        <div class="alert alert-<%= key %>">
```
jacquesporveau: Kind of weird to have a separate flash div for each flash. It might be nicer to organize this as a list or something.
Something like
```
flash-div
  ul
    flash.each do 
      li - flash-msg
      button
      /li
    end
  /ul
/div
```

jacquesporveau (commented):
Looking good! Lets get that diff cleaned up as mentioned above.

app/models/bet.rb
```diff
@@ -3,6 +3,10 @@ class Bet < ApplicationRecord
   has_many :user_bets
   has_many :options, through: :user_bets, source: :bet_option
 
+  validates :expires_at, presence: true
+  validates :creator_id, presence: true
```
isaacfreeman: You might want to add a migration to set database constraints here as well Rails validations
```change_column_null :table_name, :column_name, false```

File: app/validators/expiry_date_validator.rb (hunk 1 above)
Comment (isaacfreeman) [reply app/validators/expiry_date_validator.rb]: If you're like me and enjoy having the ? at the end to indicate that the result is a boolean, you can also do `record.expires_at.present?`.

File: app/views/layouts/application.html.erb [hunk 4]
```diff
@@ -9,6 +9,20 @@
   </head>
 
   <body>
-    <%= yield %>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery-datetimepicker/2.5.4/build/jquery.datetimepicker.full.min.js"></script>
+
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/jquery-datetimepicker/2.5.4/build/jquery.datetimepicker.min.css" />
+    <link href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet">
```
isaacfreeman: Any reason JS and stylesheets are loaded in `<body>` instead of `<head>`?

app/validators/expiry_date_validator.rb
qr8r: You should always prefer `Date`, `DateTime`, `Time` `#current` over `#now`. You should read the docs to understand why.

Be careful when writing logic statements. If you can avoid double negatives, you should. 
`!record.expires_at.nil?` can just be `record.expires_at` since in ruby objects are truthy.

app/controllers/bets_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+    @bet = Bet.new
+  end
+
+  def create
+    @bet = Bet.new(bet_params)
+    if @bet.save
+      flash[:success] = 'Bet has been created'
+      redirect_to root_path
+    else
+      @bet.errors.full_messages.each do |message|
```
qr8r: Another way to do this is `@bet.errors.full_messages.to_sentence` will join all of your errors into a human readable string

qr8r (changes requested):
Careful when structuring your PRs.  This one is a little all over the place. 

Overall good content though. A few tweaks and then Im happy to merge

isaacfreeman (approved):

app/validators/expiry_date_validator.rb
```diff
@@ -0,0 +1,10 @@
+class ExpiryDateValidator < ActiveModel::Validator
+  def validate(record)
+    if record.expires_at && record.expires_at <= DateTime.now
```
qr8r: Dont use Time.now. always use Time.current

You should look up the differrence

qr8r (changes requested):

app/validators/expiry_date_validator.rb
ghost: Sorry, I zoned out and forgot to change this one. Should be fixed now, thanks!

app/validators/expiry_date_validator.rb
```diff
@@ -0,0 +1,10 @@
+class ExpiryDateValidator < ActiveModel::Validator
+  def validate(record)
+    if record.expires_at && record.expires_at <= DateTime.current
+      record.errors[:expires_at] << 'date must be in the future'
+      false
+    else
+      true
+    end
+  end
+end
```
jacquesporveau: Doesn't look like this is being tested anywhere. Personally I don't think we need to return true or false here.
I just did some work on this myself and I think we can refactor to something like 
```
return if record.expires_at && record.expires_at <= DateTime.current
record.errors[:expires_at] << 'blehblehbleh'
```

Refrence: http://www.rails-dev.com/custom-validators-in-ruby-on-rails-4

If this was being tested somewhere it would be really easy to tell if this refactor works. ( I think it does ).

qr8r: Agreed. having read the spec (http://guides.rubyonrails.org/active_record_validations.html#custom-validators) this method doesnt need to return true or false.

ghost: Thanks for catching this one, hopefully I fixed it.

qr8r (approved):

PR #33, opened 2017-03-02
Add ability to bet on multiple options

ghost:
Dependent on #28 .
Added private helper functions to "Bet" for "resolve" method.
Provided rspec for the updated "resolve" method in "Bet".
With this adjustment, users will be able to place bets on multiple options.

### DIFF
--- a/app/models/bet.rb
+++ b/app/models/bet.rb
@@ -7,12 +7,8 @@ class Bet < ApplicationRecord
     winning_option.update!(winner: true)
     debt_relationships = user_bets.winners.to_a.product(user_bets.losers.to_a)
     debt_relationships.map do |(won_bet, lost_bet)|
-      Debt.create(
-        creditor: won_bet.user,
-        debtor: lost_bet.user,
-        amount:  amount_owed(lost_bet, won_bet)
-      )
-    end
+      resolve_debt(lost_bet, won_bet) unless lost_bet.user == won_bet.user
+    end.compact
   end
 
   def winner_total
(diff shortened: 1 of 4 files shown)


ghost:
I just realized that the "update_debt" method looks through the whole Debt database and not the debts we are creating at the moment, which might ruin the initial idea of binding debts to specific bets. I will try to resolve it in this PR.

app/models/bet.rb
```diff
@@ -27,4 +23,28 @@ def amount_owed(lost_bet, won_bet)
   def bet_total
     user_bets.pluck(:amount_bet).sum
   end
+
+  private
+
+  def resolve_debt(lost_bet, won_bet)
+    existing_debt = Debt.find_by(debtor: lost_bet.user, creditor: won_bet.user)
+    if existing_debt
+      update_debt(existing_debt, lost_bet, won_bet)
```
qr8r, line 32: I kind of feel like this is out of scope for this PR ...... I really dont want to worry about duplcated creditor -> debtor debts yet

app/models/bet.rb
```diff
@@ -27,4 +23,28 @@ def amount_owed(lost_bet, won_bet)
...
+
+  def create_debt(lost_bet, won_bet)
+    Debt.create(
+      debtor: lost_bet.user,
+      creditor: won_bet.user,
+      amount:  amount_owed(lost_bet, won_bet)
+    )
+  end
+
+  def update_debt(debt, lost_bet, won_bet)
```
qr8r, line 46: This is called feature creep. Its funtionality that was not specified in any story and should not have been implemented (yet)

qr8r (changes requested):

spec/models/bet_option_spec.rb
```diff
@@ -48,9 +53,14 @@
     end
 
     context 'New England wins' do
-      let(:option1) { BetOption.create(option_text: 'Atlanta wins') }
+      let(:option1) do
+        BetOption.create(option_text: 'Atlanta wins',
+                         bet: bet1)
```
jacquesporveau, line 58: In your bet_spec.rb file you are one lining a bet option with two parameters. Pick one syntax style and roll with it.

qr8r: Yeah, This might be an artifact of the overly strict rubocop

qr8r:
Jenkins test this please

qr8r:
Closing; No story and tests fail

PR #34, opened 2017-03-07
Add ability to view all bets

ghost:
Bets are displayed at /bets 25/page by default.
A page can be accessed manually by adding ?page=# to /bets.
Added custom Paginator module to provide user with ability to change pages with buttons.

### DIFF
--- a/app/helpers/sessions_helper.rb
+++ b/app/helpers/sessions_helper.rb
@@ -1,6 +1,7 @@
 module SessionsHelper
   def log_in(user)
     session[:user_id] = user.id
+    current_user
   end
 
   def log_out
--- a/spec/helpers/sessions_helper_spec.rb
+++ b/spec/helpers/sessions_helper_spec.rb
@@ -1,0 +1,29 @@
+require 'rails_helper'
+
+RSpec.describe SessionsHelper, type: :helper do
+  let(:user) do
+    User.create(username: 'Bob', email: 'bob@mail.com', password: '123')
+  end
+
+  describe '#log_in' do
+    subject { log_in(user) }
+    it { is_expected.to eq user }
+  end
+
+  describe '#current_user' do
+    subject { current_user }
+    before { log_in(user) }
+    it { is_expected.to eq user }
+  end
+
+  describe '#log_out' do
+    subject { log_out }
+    before { log_in(user) }
+    it 'changes session[:user_id] to nil' do
+      expect { subject }.to change { session[:user_id] }.to(nil)
+    end
+    it 'changes the current user to nil' do
+      expect { subject }.to change { @current_user }.from(@current_user).to(nil)
+    end
+  end
+end
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -55,4 +55,5 @@ RSpec.configure do |config|
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
   config.include SessionsHelper, type: :controller
+  config.include SessionsHelper, type: :helper
 end


app/views/bets/index.html.erb
```diff
@@ -5,5 +5,9 @@
       <h4>This page will display all the existing bets (25/page) and allow the user to use custom search criteria.</h4>
     </div>
   </div>
-  <%= render 'bets/display_bets' %>
+  <div class="d-flex flex-column">
+    <div class="p-2">
+      <%= render 'bets/display_bets', bets: @bets %>
```
qr8r: Nice! I really like that you pass the instance variable into the partial. This makes the partial way easier to reuse elsewhere.

For reference, the @bets instance variable is still available within the partial, but its bad practice to use instance variables in partials.

Just as as reminder that rails will start looking for partials in the current view directory. So in this case 
`<%= render 'display_bets', bets: @bets %>` would also have worked, since the index view and display bets are in the same view directory.


Another option available to you here is that Rails can infer which view to render from AR models. Making it possible to write

`<%= render @bets %>` 

This will look for a partial called `/bets/_bet.html.erb` and it will automatically loop over ever item in the @bets collection and render 1 partial for each bet. 

http://guides.rubyonrails.org/layouts_and_rendering.html#rendering-collections

app/views/bets/_display_bets.html.erb
```diff
@@ -5,10 +5,22 @@
...
+          <th class="col-md-2">Expires at</th>
           <th class="col-md-2">Total amount bet</th>
-          <th class="col-md-6">Bet description</th>
+          <th class="col-md-4">Bet description</th>
         </tr>
       </thead>
+      <tbody>
+        <% bets.each do |bet| %>
+          <tr>
+            <td class="col-md-2"><%= bet.creator_id %></td>
```
qr8r: another cool rails trick is the `content_tag` view helper. Its not always the right choice, but its a cool 'ruby-esque' way of creating tags.

`<%= content_tag :td, bet.creator_id, class: 'col-md-2' %>`

http://apidock.com/rails/ActionView/Helpers/TagHelper/content_tag

app/controllers/bets_controller.rb
```diff
@@ -1,6 +1,9 @@
 class BetsController < ApplicationController
   def index
-    @bets = Bet.all
+    @items_per_page = 25
+    @current_page = params[:page] || '1'
+    @bets = Bet.offset(@items_per_page * (@current_page.to_i - 1))
```
qr8r: This is :+1: for now,
But we might want to move this into some kind of helper if we ever want to paginate any other resources. That way we can share this code between controllers

app/views/shared/_paginator.html.erb
```diff
@@ -0,0 +1,37 @@
+<div class="paginator">
+  <ul class="pagination">
```
qr8r: Nice! I like the idea of using a list here. But I think if we are going to do so we might want to use an `ol`.
This is a nitpicky comment... but a list of ordered numbers strikes me as list where order matters.

app/views/shared/_paginator.html.erb
```diff
@@ -0,0 +1,37 @@
...
+    <li class="<%= previous_status %>">
```
qr8r: This is a really good place to use the `content_tag` syntax. Mixing ruby and erb can be messy and hard to read.

```ruby
<%= content_tag :li, class: previous_state do %>
   ...
<% end %>
```

is just a little nice IMO

app/views/shared/_paginator.html.erb
```diff
@@ -0,0 +1,37 @@
...
+      <% end %>
+    </li>
+    <li class="<%= previous_status %>">
+      <%= link_to bets_path(page: current_page.to_i - 1), class: 'page-link' do %>
+        <i class="fa fa-angle-left"></i>
+      <% end %>
+    </li>
+
+    <% blocks.each do |i|%>
+      <% status = '' %>
```
qr8r: Technically this is not needed. Unlike some other languages, in ruby a variable initalized within an if block will be available outside the if statement, even if it doesnt run.

```ruby
if 1 == 3
  thing = 6
end 

puts thing.inspect # => nil
```

Also, since this is just one big block of ruby, you dont need to wrap each line in `<% %>` 

you could write it:
 
```ruby
<%
  if stuff 
    ...
  end
%>
```

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,81 @@
+module Utilities
```
qr8r: Woot! This is called a service class. Is a Ruby class that is not tied to the database used to do some heavy computational lifting.

Service classes are great and really help clean up your code elsewhere like in your view above
:+1: :100:  for this pattern

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,81 @@
...
+  module Paginator
+    def paginator(max_page_blocks: 9,
+                  items_per_page: 5,
+                  items: number_of_all_bets,
+                  current_page:)
+      lp = last_page(items, items_per_page)
+      ptd = pages_to_display(current_page.to_i, max_page_blocks.to_i, lp.to_i)
+      render partial: 'shared/paginator',
```
qr8r: This is old rails syntax.

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,81 @@
...
+                       current_page: current_page,
+                       previous_status: previous_status(current_page),
+                       next_status: next_status(current_page, lp),
+                       last_page: lp }
+    end
+
+    private
+
+    def last_page(number_of_items, items_per_page)
+      (number_of_items.to_f / items_per_page).ceil.to_s
```
qr8r: Why return a string?

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,81 @@
+module Utilities
+  module Paginator
+    def paginator(max_page_blocks: 9,
+                  items_per_page: 5,
+                  items: number_of_all_bets,
+                  current_page:)
+      lp = last_page(items, items_per_page)
+      ptd = pages_to_display(current_page.to_i, max_page_blocks.to_i, lp.to_i)
```
qr8r: :( abrvs mk it vry hrd 2 rd

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+    private
+
+    def last_page(number_of_items, items_per_page)
+      (number_of_items.to_f / items_per_page).ceil.to_s
+    end
+
+    def previous_status(current_page)
+      current_page == '1' ? 'page-item disabled' : 'page-item'
```
qr8r: Might not be a bad idea to cast this to an int.

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+    def next_status(current_page, last_page)
+      current_page == last_page ? 'page-item disabled' : 'page-item'
+    end
+
+    def pages_to_display(current_page, mpb, number_of_pages)
```
qr8r: I like all of this, but I think we can simplify it. 

Id like to see some test cases for this module, then we can start to refactor it to make it just a little nicer to read through

app/controllers/bets_controller.rb
```diff
@@ -1,9 +1,13 @@
 class BetsController < ApplicationController
+  include Utilities::Paginator
+  helper_method :paginator
```
qr8r: Might be able to just call `helper Utilities::Paginator` if we make paginator the only public api for the Paginator module. 
Not 100% sure thats the right move...

app/controllers/bets_controller.rb
```diff
@@ -1,9 +1,13 @@
...
+
   def index
     @items_per_page = 25
     @current_page = params[:page] || '1'
     @bets = Bet.offset(@items_per_page * (@current_page.to_i - 1))
                .limit(@items_per_page)
+    @number_of_all_bets = Bet.all.count
```
qr8r: I think I would like to see some of the logic from lines 8 - 10 moved into the paginator somehow.

In general we try and avoid computation in controllers. They are really only for loading resources and rending things

app/views/bets/index.html.erb
```diff
@@ -9,5 +9,14 @@
     <div class="p-2">
       <%= render 'bets/display_bets', bets: @bets %>
     </div>
+    <div class="mt-auto p-2">
+      <div class="row footer">
+        <div class="col col-md-12">
+          <%= paginator items: @number_of_all_bets,
```
qr8r: :+1: :100: This was exactly what I was imagining when we talked about this

qr8r (changes requested):
I've requested changes because I think this can be improved.

But Im super happy with the size of these commits, and the size of the PR. 

I am also happy with the direction you chose to implement this feature. It encapsulates the pagination feature into a cohesive service class, and the appropriate controller / view changes are in appropriate commits as well.

I would like to see specs for the Paginator Utility class, and a note for future improvement would be that the Paginator could probably have been split up over a few  commits that implement 1 method + 1 spec. But as it is I still think that the commit you have is a reasonable size (although missing specs)

app/views/bets/_display_bets.html.erb
```diff
@@ -5,10 +5,22 @@
...
+            <td class="col-md-2"><%= bet.created_at %></td>
+            <td class="col-md-2"><%= bet.expires_at %></td>
+            <td class="col-md-2"><%= bet.bet_total %></td>
+            <td class="col-md-4"><%= bet.bet %></td>
```
jacquesporveau: Reading this code I have no idea what `bet.bet` is. I feel like the naming could be more clear somewhere in there.

ghost: My bad, I should have named it bet_text or something along the lines.

app/views/bets/_display_bets.html.erb
```diff
@@ -5,10 +5,22 @@
...
           <th class="col-md-2">Created By</th>
           <th class="col-md-2">Created at</th>
+          <th class="col-md-2">Expires at</th>
           <th class="col-md-2">Total amount bet</th>
-          <th class="col-md-6">Bet description</th>
+          <th class="col-md-4">Bet description</th>
         </tr>
       </thead>
+      <tbody>
+        <% bets.each do |bet| %>
```
qr8r: Is this partial still used?

app/views/shared/_paginator.html.erb
```diff
@@ -0,0 +1,52 @@
+<div class="paginator">
+  <ol class="pagination">
+    <%=
+      content_tag :li, class: previous_status do
+        link_to bets_path(page: '1'), class: 'page-link' do
+          content_tag :i, class: 'fa fa-angle-double-left' do
```
qr8r: Empty block

app/views/shared/_paginator.html.erb
```diff
@@ -0,0 +1,52 @@
...
+          end
+        end
+      end
+    %>
+    <%=
+      content_tag :li, class: previous_status do
+        link_to bets_path(page: current_page - 1), class: 'page-link' do
+          content_tag :i, class: 'fa fa-angle-left' do
```
qr8r: Same

app/views/shared/_paginator.html.erb
```diff
@@ -0,0 +1,52 @@
...
+          end
+        end
+      end
+    %>
+
+    <% blocks.each do |i|%>
```
qr8r: blocks could be a partial

app/views/shared/_paginator.html.erb
```diff
@@ -0,0 +1,52 @@
...
+      <%
+        if i == current_page
```
qr8r: This is too much logic for a partial, It should live somewhere else.... also single letter variable

app/views/shared/_paginator.html.erb
```diff
@@ -0,0 +1,52 @@
...
+      content_tag :li, class: "page-item #{status}" do
+        link_to i, bets_path(page: i), class: 'page-link'
+      end
+    %>
+    <% end %>
+
+    <%=
+      content_tag :li, class: next_status do
+        link_to bets_path(page: current_page + 1), class: 'page-link' do
+          content_tag :i, class: 'fa fa-angle-right' do
```
qr8r: Empty block

app/views/shared/_paginator.html.erb
```diff
@@ -0,0 +1,52 @@
...
+          end
+        end
+      end
+    %>
+    <%=
+      content_tag :li, class: next_status do
+        link_to bets_path(page: last_page), class: 'page-link' do
+          content_tag :i, class: 'fa fa-angle-double-right' do
```
qr8r: empty block

spec/modules/paginator_spec.rb
```diff
@@ -0,0 +1,93 @@
+require 'rails_helper'
+
+RSpec.describe Utilities::Paginator, type: :module do
+  let!(:dummy) { Class.new { extend Utilities::Paginator } }
```
qr8r: This is cute.

This might be helpful though https://relishapp.com/rspec/rspec-rails/v/3-5/docs/helper-specs/helper-spec

qr8r (changes requested):

qr8r:
Jenkins test this please

qr8r:
Jenkins test this please

qr8r:
Jenkins test this please

config/application.rb
```diff
@@ -11,5 +11,6 @@ class Application < Rails::Application
     # Settings in config/environments/* take precedence over those specified here.
```
qr8r: @YarekTenko The commit message for this change is wrong. 

Autoload does not  automatically load constants upon the
application start.

Please read http://guides.rubyonrails.org/autoloading_and_reloading_constants.html and make sure to understand what autoloading does

app/views/bets/index.html.erb
```diff
@@ -5,5 +5,16 @@
       <h4>This page will display all the existing bets (25/page) and allow the user to use custom search criteria.</h4>
     </div>
   </div>
-  <%= render 'bets/display_bets' %>
+  <div class="d-flex flex-column">
+    <div class="p-2">
+      <%= render 'display_bets', bets: @bets %>
```
qr8r: The _bet partial should be included in this commit

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,85 @@
+module Utilities
+  module Paginator
+    def items_to_display(item:, items_per_page:)
+      @items_per_page = items_per_page || 5
```
qr8r: Keyword arguments can have a default values instead of using this conditional statement

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,85 @@
...
+      @current_page = params[:page] ? params[:page].to_i : 1
```
qr8r: I dont like this class looking at the params. I think the current page should be passed in

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,85 @@
...
+      @number_of_items = item.all.count
```
qr8r: Active record has a method to get the number of DB records for a class. The only reason to perform this computation in ruby would be for performance (but you have not implemented the rest of the changes needed to actually optimise for this)

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,85 @@
+module Utilities
+  module Paginator
+    def items_to_display(item:, items_per_page:)
```
qr8r: Item is misleading.... It makes me think this will be an instance, instead of a class

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,85 @@
...
+                                                @last_page),
+                       current_page: @current_page,
+                       previous_status: previous_status,
+                       next_status: next_status,
+                       last_page: @last_page }
+    end
+
+    private
+
+    def previous_status
```
qr8r: This method could be more semantic. It returns the css class name for the "previous" link

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,85 @@
...
+      @current_page == 1 ? 'page-item disabled' : 'page-item'
+    end
+
+    def next_status
```
qr8r: Same same

qr8r: Same same

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,85 @@
...
+      @current_page == @last_page ? 'page-item disabled' : 'page-item'
+    end
+
+    def pages_to_display(current_page, mpb, number_of_pages)
```
qr8r: This method needs to be much more readable

spec/modules/paginator_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      let(:current_page) { 1 }
+      let(:number_of_pages) { 5 }
+      let(:expected_range_of_pages) { [1, 2, 3, 4, 5] }
+
+      it 'displays correct range of pages' do
+        expect(pages_to_display).to eq expected_range_of_pages
+      end
+    end
+
+    context 'Current page is in the beginning and there are more pages towards the end of the pagination list' do
```
qr8r: These are really good specs :) Good job

qr8r (changes requested):

cbrunsdon:
jenkins test this please

app/views/bets/_display_bets.html.erb
ghost: Thanks for the feedback! Yes, it's the one that displays the bets on the screen, depending on how many have been passed to bets

PR #35, opened 2017-03-08
Users can view a particular bet

ghost:
Users can view particular bets by accessing /bets/:id.

### DIFF
--- a/app/models/bet.rb
+++ b/app/models/bet.rb
@@ -2,6 +2,7 @@ class Bet < ApplicationRecord
   has_many :users, through: :user_bets
   has_many :user_bets
   has_many :options, class_name: 'BetOption'
+  belongs_to :creator, class_name: 'User'
 
   validates :expires_at, presence: true
   validates :creator_id, presence: true
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -3,6 +3,7 @@ class User < ApplicationRecord
   has_many :user_bets
   has_many :debts,   class_name: 'Debt', foreign_key: :debtor_id
   has_many :credits, class_name: 'Debt', foreign_key: :creditor_id
+  has_many :created_bets, class_name: 'Bet'
 
   has_secure_password
 end
--- a/app/views/bets/show.html.erb
+++ b/app/views/bets/show.html.erb
@@ -1,49 +1,16 @@
-<div class="container">
-  <hr>
-  <div class="row">
-    <div class="col col-lg-12">
-      <h2>Viewing bet: <%= @bet.id %></h2>
-    </div>
-  </div>
-  <hr>
-  <div class="row">
-    <div class="col col-lg-4">
-      <h4>ID</h4>
-    </div>
-    <div class="col col-lg-8">
-      <h4><%= @bet.id %></h4>
-    </div>
-  </div>
-  <div class="row">
-    <div class="col col-lg-4">
-      <h4>Status</h4>
-    </div>
-    <div class="col col-lg-8">
-      <h4><%= @bet.status %></h4>
-    </div>
-  </div>
-  <div class="row">
-    <div class="col col-lg-4">
-      <h4>Creator ID</h4>
-    </div>
-    <div class="col col-lg-8">
-      <h4><%= @bet.creator_id %></h4>
-    </div>
-  </div>
-  <div class="row">
-    <div class="col col-lg-4">
-      <h4>Expires at</h4>
-    </div>
-    <div class="col col-lg-8">
-      <h4><%= @bet.expires_at %></h4>
-    </div>
-  </div>
-  <div class="row">
-    <div class="col col-lg-4">
-      <h4>Text</h4>
-    </div>
-    <div class="col col-lg-8">
-      <h4><%= @bet.bet %></h4>
+<hr>
+<div class="row">
+  <div class="col col-lg-6 offset-lg-3">
+    <div class="card text-center">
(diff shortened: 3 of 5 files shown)


ghost:
Minor adjustment: now the bets and users are displayed in for of a table instead of card blocks.

app/models/bet.rb
```diff
@@ -2,6 +2,7 @@ class Bet < ApplicationRecord
   has_many :users, through: :user_bets
   has_many :user_bets
   has_many :options, class_name: 'BetOption'
+  belongs_to :creator, class_name: 'User', foreign_key: :creator_id
```
qr8r: Is the foreign_key declaration required here?

app/models/user.rb
```diff
@@ -6,4 +6,8 @@ class User < ApplicationRecord
   has_many :created_bets, class_name: 'Bet', foreign_key: :creator_id
 
   has_secure_password
+
+  def waged_on(bet_option)
+    user_bets.where(bet_option: bet_option).pluck(:amount_bet).sum
```
qr8r: Can a user bet on the same option more than once?

app/views/bets/show.html.erb
```diff
@@ -1,49 +1,73 @@
-<div class="container">
-  <hr>
-  <div class="row">
-    <div class="col col-lg-12">
-      <h2>Viewing bet: <%= @bet.id %></h2>
+<!-- Bet -->
```
qr8r: Write code so good that doesnt need comments ;)

app/views/bets/show.html.erb
```diff
@@ -1,49 +1,73 @@
...
+      <div class="card-header text-muted">
+        <h4><%= @bet.creator.username %></h4>
+      </div>
+
+      <div class="class-block">
+        <p><h5 class="card-text"><%= @bet.bet %></h5></p>
+      </div>
+
+      <div class="card-footer text-danger">
+        Expires on: <%= @bet.expires_at.strftime('%A, %B %d %Y, %l:%M %P') %>
```
qr8r: :+1: `#strftime`

app/views/bets/show.html.erb
```diff
@@ -1,49 +1,73 @@
...
+</div>
+
+<!-- Users and their wagers -->
+<hr>
+<div class="row">
+  <div class="col col-lg-6 offset-lg-3">
+    <div class="card">
+      <div class="card-header text-muted">
+        <div class="row">
+          <div class="col col-lg-4">User</div>
```
qr8r, line 24: divitis

app/views/bets/show.html.erb
```diff
@@ -1,49 +1,73 @@
...
+          <div class="col col-lg-4">Option #</div>
+          <div class="col col-lg-4">Amount</div>
+        </div>
+      </div>
+      <div class="card-block">
+        <table class="table table-hover">
+          <% @bet.user_bets.each do |ub| %>
```
qr8r: Might be a good opportunity for a `<%= render @bets.user_bets %>`

qr8r (approved):
Minor comments, but overall :1st_place_medal:

qr8r (changes requested):
Meant to request changes earlier

app/views/bets/show.html.erb
```diff
@@ -1,49 +1,73 @@
...
+          <div class="col col-lg-4">Option total</div>
+        </div>
+      </div>
+      <div class="card-block">
+        <table class="table table-hover">
+          <% @bet.options.each_with_index do |option, i| %>
+            <tr class="row">
+              <td class="col col-lg-4"><%= i + 1 %></td>
+              <td class="col col-lg-4"><%= option.option_text %></td>
+              <td class="col col-lg-4">$<%= option.option_total %></td>
```
jacquesporveau: Mmmmmm the naming of the BetOption model, and option_total method are kind of weird. You shouldn't need to specify that you are looking for the "option" total when you are calling it on an object called option. Just `option.total`. 

I realize that this isn't really related to this PR but it's just something to keep in mind.

qr8r: :+1: Naming is hard, but I do agree. We dont need to change this now. But definitely something to keep in mind for the future

qr8r:
Jenkins test this please

cbrunsdon:
jenkins test this please

app/views/bets/show.html.erb
```diff
@@ -1,49 +1,16 @@
...
-      <h4>Text</h4>
-    </div>
-    <div class="col col-lg-8">
-      <h4><%= @bet.bet %></h4>
+<hr>
+<div class="row">
+  <div class="col col-lg-6 offset-lg-3">
+    <div class="card text-center">
+      <div class="card-header text-muted">
+        <h4><%= @bet.creator.username %></h4>
```
jacquesporveau, line 6: Is creator a required association on bet? What would happen here if a bets creator was `nil`? Same thing with username, I am not seeing any validations on a users username. What if they didn't have a user name?

ghost: Yeah, basically, creator can't be nil, otherwise the Bet won't be created.
Username is not validated yet, but there is a separate story for that, so I decided not to implement it here.
Thanks for the feedback!

qr8r (approved):

app/views/bets/show.html.erb
qr8r: We lost the fight to divitis.... but the functionality meets the story

PR #40, opened 2017-03-16
Fixing rspecs for bet and bet_option

ghost:
Added required fields to test variables that are now validated upon creation, which caused them to break the rspec.

### DIFF
--- a/spec/models/bet_option_spec.rb
+++ b/spec/models/bet_option_spec.rb
@@ -1,18 +1,19 @@
 require 'rails_helper'
 
 RSpec.describe BetOption, type: :model do
-  let!(:bob) do
+  let(:bob) do
     User.create(username: 'Bob', email: 'bob@mail.com', password: '123')
   end
-  let!(:peter) do
+  let(:peter) do
     User.create(username: 'Peter', email: 'peter@mail.com', password: 'pete1')
   end
-  let!(:jack) do
+  let(:jack) do
     User.create(username: 'Jack', email: 'jack@mail.com', password: 'notjack')
   end
 
   let(:bet1) do
-    Bet.create(bet: 'Superbowl: Atlanta vs New England, place your bets!')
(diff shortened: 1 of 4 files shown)


spec/models/bet_spec.rb
```diff
@@ -11,11 +11,14 @@
     User.create(username: 'Jack', email: 'jack@mail.com', password: 'notjack')
   end
 
-  let(:bet1) do
-    Bet.create(bet: 'Superbowl: Atlanta vs New England, place your bets!')
+  let!(:bet1) do
+    Bet.create(bet: 'Superbowl: Atlanta vs New England, place your bets!',
+               expires_at: DateTime.current + 1, creator_id: bob.id)
```
qr8r: Rails has some fun helpers for playing with dates 

http://api.rubyonrails.org/classes/Date.html

spec/models/bet_spec.rb
```diff
@@ -11,11 +11,14 @@
...
+  end
+  let!(:option1) { BetOption.create(option_text: 'Atlanta wins', bet: bet1) }
```
qr8r: I dont think these need to be bang methods.... Can you explain why you think you need them?

spec/models/bet_option_spec.rb
```diff
@@ -11,8 +11,9 @@
     User.create(username: 'Jack', email: 'jack@mail.com', password: 'notjack')
   end
 
-  let(:bet1) do
-    Bet.create(bet: 'Superbowl: Atlanta vs New England, place your bets!')
+  let!(:bet1) do
+    Bet.create(bet: 'Superbowl: Atlanta vs New England, place your bets!',
+               expires_at: DateTime.current + 1, creator_id: bob.id)
```
qr8r: Same comments as your previous test

qr8r (changes requested):

cbrunsdon:
jenkins test this please

qr8r (approved):

PR #44, opened 2017-03-20
Add rspec for bets_controller

ghost:
Added rspec for controllers/bets_controller.rb

### DIFF
--- a/spec/controllers/bets_controller_spec.rb
+++ b/spec/controllers/bets_controller_spec.rb
@@ -1,0 +1,59 @@
+require 'rails_helper'
+
+RSpec.describe BetsController, type: :controller do
+  describe 'GET #index' do
+    subject { get :index }
+    it { is_expected.to be_ok }
+  end
+
+  describe 'GET #show' do
+    subject { get :show, params: { id: bet.id } }
+    let(:user) do
+      User.create(username: 'Bob', email: 'bob@mail.com', password: '123')
+    end
+    let(:bet) do
+      Bet.create(bet: 'Bet text',
+                 expires_at: DateTime.tomorrow,
+                 creator_id: user.id)
+    end
+    it { is_expected.to be_ok }
+  end
+
+  describe 'GET #new' do
+    subject { get :new }
+    it { is_expected.to be_ok }
+  end
+
+  describe 'POST #create' do
+    subject(:post_create) { post :create, params: { bet: bet_params } }
+    let(:user) do
+      User.create(username: 'Bob', email: 'bob@mail.com', password: '123')
+    end
+    context 'with valid params' do
+      let(:bet_params) do
+        { bet: 'Bet text',
+          expires_at: DateTime.tomorrow,
+          creator_id: user.id }
+      end
+      let(:created_bet) { Bet.last }
+      it 'creates a new bet in the database' do
+        expect { post_create }.to change(Bet, :count).by(1)
+      end
+      it { is_expected.to redirect_to(created_bet) }
+    end
+    context 'with invalid params' do
+      let(:bet_params) do
+        { bet: 'Invalid Bet',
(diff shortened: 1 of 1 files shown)


spec/controllers/bets_controller_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    let(:invalid_params) do
+      { bet: 'Invalid Bet',
+        expires_at: DateTime.yesterday,
+        creator_id: user.id }
+    end
+    subject { post :create, params: { bet: bet_params } }
+    context 'with valid params' do
+      let(:bet_params) { valid_params }
+      let(:created_bet) { Bet.last }
+      it { expect { subject }.to change(Bet, :count).by(1) }
```
swively: when you aren't able to write `it { is_expected ... ` in a way that makes sense when reading it back, it is preferrable to write it out on multiplte lines:
```
it 'changes the bet count' do
  expect(subject).to change(etc)
end
```

This is also a situation where naming your subject can helpful
i.e `subject(:get_new) { get :new }` lets you write things like `expect(get_new).to do_something_cool`

swively (dismissed):

spec/controllers/bets_controller_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      { bet: 'Bet text',
+        expires_at: DateTime.tomorrow,
+        creator_id: user.id }
+    end
+    let(:invalid_params) do
+      { bet: 'Invalid Bet',
+        expires_at: DateTime.yesterday,
+        creator_id: user.id }
+    end
+    subject { post :create, params: { bet: bet_params } }
```
qr8r: This is a style comment, but I prefer the subject to be the first thing in the test, so its easy to find exactly what is being tested.

```ruby 
describe 'thing' do
  subject { ... }
end
```

spec/controllers/bets_controller_spec.rb
ghost: Thanks, Mike, hopefully I fixed it now.

cbrunsdon:
jenkins test this please

spec/controllers/bets_controller_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+        expires_at: DateTime.tomorrow,
+        creator_id: user.id }
+    end
+    let(:invalid_params) do
+      { bet: 'Invalid Bet',
+        expires_at: DateTime.yesterday,
+        creator_id: user.id }
+    end
+    context 'with valid params' do
+      let(:bet_params) { valid_params }
```
qr8r: Why not define this directly here

qr8r (changes requested):

spec/controllers/bets_controller_spec.rb
ghost: Thanks! Should be fixed now.

qr8r (approved):

spec/controllers/bets_controller_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    it { is_expected.to be_ok }
+  end
+
+  describe 'GET #new' do
+    subject { get :new }
+    it { is_expected.to be_ok }
+  end
+
+  describe 'POST #create' do
+    subject(:post_create) { post :create, params: { bet: bet_params } }
```
jacquesporveau, line 28: In all your other specs above this you aren't overwriting the `subject`'s name but in this case you are. Any reasoning behind that? I would like to see a more consistent method being taken.

jacquesporveau: Yarek has told me in person the reason he made these changes and they make sense to me.

jacquesporveau:
This branch is out of date with the base branch. Please update.

qr8r:
Needs rebase

PR #48, opened 2017-03-20
Change "status" to "active" for Bet

ghost:
-Changed the type from integer to boolean and set it to be equal to true
by default (when the bet is created).


qr8r (changes requested):
There are 2 places in your code where you still refer to this attribute as status. 

This commit breaks your app

PR #49, opened 2017-03-20
Add rspec for bets_helper

ghost:
Dependent on #35 , #48 
Added rspec for helpers/bets_helper.rb

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -66,3 +66,4 @@ end
 
 gem "font-awesome-rails"
 
+gem 'timecop'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -191,6 +191,7 @@ GEM
     thor (0.19.4)
     thread_safe (0.3.5)
     tilt (2.0.6)
+    timecop (0.8.1)
     turbolinks (5.0.1)
       turbolinks-source (~> 5)
     turbolinks-source (5.0.0)
@@ -232,6 +233,7 @@ DEPENDENCIES
   spring
   spring-watcher-listen (~> 2.0.0)
   sqlite3
+  timecop
   turbolinks (~> 5)
   tzinfo-data
   uglifier (>= 1.3.0)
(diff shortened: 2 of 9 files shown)


app/helpers/bets_helper.rb
```diff
@@ -0,0 +1,9 @@
+module BetsHelper
+  def creator?
+    current_user == @bet.creator
+  end
+
+  def resolved?
```
fredericboivin: 1) Name is confusing, since the field is called `active`, this should probably be `active?` which is already provided by default by Rails/ActiveRecord for boolean fields, removing the need for this helper as you can directly use `active?`

2) This would rely on `@bet` being defined somewhere else as an instance variable, which is confusing and unreliable, so instead a bet should be provided as a parameter for scenarios like this

app/helpers/bets_helper.rb
```diff
@@ -0,0 +1,9 @@
+module BetsHelper
+  def creator?
+    current_user == @bet.creator
```
fredericboivin: @bet should a parameter for the reasons mentioned below

1) If @bet is not defined, this would raise an Internal Server Error
2) Assuming that the variable is defined somewhere is an unsafe assumption
3) There is no guarantee that this the data you expect as different parts of the code could
set and override each other @bet variable

This should also be a instance method on the Bet model or the User model (for example `@bet.created_by?(user)`)

fredericboivin (changes requested):

app/helpers/bets_helper.rb
dangerdogz: Generally, I don't think an helper is warranted in this case. The helpers should be instance methods on the Bet model

fredericboivin (changes requested):
Those helpers should be replaced by instance methods on the Bet model

spec/helpers/bet_helper_spec.rb
```diff
@@ -0,0 +1,60 @@
+require 'rails_helper'
+require 'pry'
```
qr8r: :crying_cat_face:

cbrunsdon:
jenkins test this please

app/helpers/bets_helper.rb
ghost: Thanks for the feedback, Fred! I'll do my best to fix it.

PR #50, opened 2017-03-20
Modify log_in method in SessionsHelper

ghost:
-Modified the log_in method in the SessionsHelper: now it automatically sets the current_user as the logged in user.
-Added rspec for helpers/sessions_helper.rb
-Added SessionHelper to rails_helper.rb for "helper" types.

### DIFF
--- a/spec/helpers/sessions_helper_spec.rb
+++ b/spec/helpers/sessions_helper_spec.rb
@@ -1,0 +1,29 @@
+require 'rails_helper'
+
+RSpec.describe SessionsHelper, type: :helper do
+  let(:user) do
+    User.create(username: 'Bob', email: 'bob@mail.com', password: '123')
+  end
+
+  describe '#log_in' do
+    subject { log_in(user) }
+    it { is_expected.to eq user }
+  end
+
+  describe '#current_user' do
+    subject { current_user }
+    before { log_in(user) }
+    it { is_expected.to eq user }
+  end
+
+  describe '#log_out' do
+    subject { log_out }
+    before { log_in(user) }
+    it 'changes session[:user_id] to nil' do
+      expect { subject }.to change { session[:user_id] }.to(nil)
+    end
+    it 'changes the current user to nil' do
+      expect { subject }.to change { @current_user }.from(@current_user).to(nil)
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


spec/helpers/sessions_helper_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'rails_helper'
+
+RSpec.describe SessionsHelper, type: :helper do
+  let(:user) do
+    User.create(username: 'Bob', email: 'bob@mail.com', password: '123')
+  end
+
+  describe '#log_in' do
+    subject { log_in(user) }
+    let(:logged_in_user) { user.id }
```
luukveenis: These `let`s make the test harder to read in my opinion, I would just:
`it { is_expected.to be user.id }`

spec/helpers/sessions_helper_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    let(:logged_in_user) { user }
+    before { log_in(user) }
+    it { is_expected.to eq logged_in_user }
+  end
+
+  describe '#log_out' do
+    subject { log_out }
+    before do
+      log_in(user)
+      current_user
```
luukveenis: I'm not sure what `current_user` does, but does it need to be called here?

spec/helpers/sessions_helper_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+    it { expect { subject }.to change { session[:user_id] } }
```
luukveenis: This example is checking more than just a simple return value, so I would prefer a better description on the example, ie:
```
it "sets the current user id in the session" do
  ...
end
```

spec/helpers/sessions_helper_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    it do
```
luukveenis: same here, a better description would be great

luukveenis (changes requested):

cbrunsdon:
jenkins test this please

spec/helpers/sessions_helper_spec.rb
ghost: Thanks for all the feedback, Luuk! I hope I made it better this time.

qr8r (approved):

luukveenis (approved):

PR #55, opened 2017-03-21
Add created_by? method for Bet

ghost:
Dependent on #35 : bet.creator can not be called until #35 is merged.
-Added new method created_by?(user) that checks if the bet was created by the passed User.
-Provided rspec

### DIFF
--- a/app/models/bet.rb
+++ b/app/models/bet.rb
@@ -34,4 +34,8 @@ class Bet < ApplicationRecord
   def bet_total
     user_bets.pluck(:amount_bet).sum
   end
+
+  def created_by?(user)
+    user == creator
+  end
 end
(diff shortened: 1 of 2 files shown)


jacquesporveau:
To me I get a bit of a code smell seeing this method. I would be curious about the situations you are intending on using it. Checking if a user is the creator seems less efficient that just calling `bet.creator` and getting the user object back.

Also your Jenkins checks are failing and this branch is out of date with master. Let's try to make sure that we always have your PRs in the closest to merge state you can provide.

jacquesporveau:
This branch is out of date with master and failing Jenkins. Please update and sleuth into the Jenkins stuff.

app/models/bet.rb
```diff
@@ -34,4 +34,8 @@ def amount_owed(lost_bet, won_bet)
   def bet_total
     user_bets.pluck(:amount_bet).sum
   end
+
+  def created_by?(user)
```
qr8r, line 38: Where is this method used?

PR #56, opened 2017-03-21
Create DebtFilter module for Bet

ghost:
Pivotal: https://www.pivotaltracker.com/story/show/141907091
-DebtFilter addresses the problem, where some of the created Debts for a
particular Bet would cross-reference each other (PersonA -$50-> PersonB
&& PersonB -$20-> PersonA) or accumulate as separate entries (PerosnA
-$50-> PersonB && PersonA -$20-> PersonB) by filtering them before
saving.
-Added DebtFilter to Bet model. Debts are now filtered before getting
saved.

### DIFF
--- a/app/models/debt_filter.rb
+++ b/app/models/debt_filter.rb
@@ -1,0 +1,56 @@
+module DebtFilter
+  def save_debts(debts)
+    filter_debts(debts).map do |debt|
+      debt.save
+      debt
+    end
+  end
+
+  private
+
+  def filter_debts(debts)
+    reverse_negatives(merge_debts(reverse_duplicates(debts)))
+  end
+
+  def unique_participants(debts)
+    uniques = []
+    debts.each do |debt|
+      next if uniques.include?([debt.creditor, debt.debtor]) ||
+              uniques.include?([debt.debtor, debt.creditor])
+      uniques << [debt.debtor, debt.creditor]
+    end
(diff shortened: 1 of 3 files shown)


app/models/debt_filter.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+  end
+
+  private
+
+  def filter_debts(debts)
+    reverse_negatives(merge_debts(reverse_duplicates(debts)))
+  end
+
+  def unique_participants(debts)
```
luukveenis, line 15: This whole method can be written in a much simpler way.
Hint: Think about the fact that you're creating an array called `uniques` and ensuring uniqueness of the things you're adding to it.

app/models/debt_filter.rb
```diff
@@ -0,0 +1,55 @@
+module DebtFilter
```
luukveenis, line 1: This module has one public method, but 5 private methods that handle combining debts between the same people. These methods also iterate over the same array many times, maybe see if you can simplify this.

luukveenis (changes requested):
I think there's room to simplify the `DebtFilter` module a lot. Instead of pointing out lots of little changes, I'd encourage you to try refactoring it.

jacquesporveau:
This branch is out of date with master. Please update.

PR #57, opened 2017-03-22
Add payed column to debts

ghost:
Will be used to indicate whether the debt has been payed or not.


db/migrate/20170322230926_add_payed_column_to_debt.rb
```diff
@@ -0,0 +1,5 @@
+class AddPayedColumnToDebt < ActiveRecord::Migration[5.0]
+  def change
+    add_column :debts, :payed, :boolean, default: false
```
jacquesporveau, line 3: I think you might mean paid?

ghost: Sh*t, thanks!

PR #58, opened 2017-03-22
Add paid column to debts

ghost:
This column will help determine which debt has been paid.


qr8r (approved):

qr8r:
As an alternative to a boolean field, this could also be a date field `:payment_date` Which gives you a little more information

PR #59, opened 2017-03-23
Add has_many Debts relationship to Bet

ghost:
This will help easily display or manipulate the Debts that are associated
with a particular Bet.

### DIFF
--- a/db/migrate/20170323164658_add_has_many_belongs_to_relationship_between_bet_and_debt.rb
+++ b/db/migrate/20170323164658_add_has_many_belongs_to_relationship_between_bet_and_debt.rb
@@ -1,0 +1,6 @@
+class AddHasManyBelongsToRelationshipBetweenBetAndDebt <
+  ActiveRecord::Migration[5.0]
+  def change
+    add_reference :debts, :bet, foreign_key: true
+  end
+end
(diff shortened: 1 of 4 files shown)


db/migrate/20170323164658_add_has_many_belongs_to_relationship_between_bet_and_debt.rb
```diff
@@ -0,0 +1,8 @@
+class AddHasManyBelongsToRelationshipBetweenBetAndDebt <
+  ActiveRecord::Migration[5.0]
+  def change
```
luukveenis, line 3: You should be able to do this with one line:
`add_reference :debts, :bet, foreign_key: true`

This creates the column, foreign key constraint, and by default adds an index as well:
http://edgeapi.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-add_reference

luukveenis (dismissed):

db/migrate/20170323164658_add_has_many_belongs_to_relationship_between_bet_and_debt.rb
ghost: Thanks, that's very useful!

jacquesporveau:
LGTM

qr8r (approved):

PR #60, opened 2017-03-27
Create Paginator Module

ghost:
Pivotal: https://www.pivotaltracker.com/story/show/140512959
Paginator Module displays X number of Model items per page and outputs a corresponding pagination panel.

### DIFF
--- a/app/views/shared/_paginator.html.erb
+++ b/app/views/shared/_paginator.html.erb
@@ -1,0 +1,38 @@
+<div class="paginator">
+  <ol class="pagination">
+    <%=
+      content_tag :li, class: previous_link_css_class do
+        link_to bets_path(page: '1'), class: 'page-link' do
+          content_tag :i, nil, class: 'fa fa-angle-double-left'
+        end
+      end
+    %>
+    <%=
+      content_tag :li, class: previous_link_css_class do
+        link_to bets_path(page: current_page - 1), class: 'page-link' do
+          content_tag :i, nil, class: 'fa fa-angle-left'
+        end
+      end
+    %>
+
+    <% blocks.each do |page|%>
+      <%= render 'shared/blocks', page: page,
+        page_item_css_class: page_item_css_class(page, current_page) %>
+    <% end %>
+
+    <%=
+      content_tag :li, class: next_link_css_class do
+        link_to bets_path(page: current_page + 1), class: 'page-link' do
+          content_tag :i, nil, class: 'fa fa-angle-right'
+        end
+      end
+    %>
+    <%=
+      content_tag :li, class: next_link_css_class do
+        link_to bets_path(page: last_page), class: 'page-link' do
+          content_tag :i, nil, class: 'fa fa-angle-double-right'
+        end
+      end
+    %>
+  </ol>
+</div>
--- a/lib/utilities/paginator.rb
+++ b/lib/utilities/paginator.rb
@@ -1,0 +1,111 @@
+module Utilities
+  module Paginator
+    ELLIPSES = '...'.freeze
+    OFFSET = 1
+    FIRST_PAGE = 1
+    DIVIDER = 2
+
+    def items_to_display(model:, items_per_page: 5, current_page: 1)
+      @items_per_page = items_per_page
+      @current_page = current_page
+      @number_of_items = model.count
+      @last_page = (@number_of_items.to_f / @items_per_page).ceil
+      model.offset(items_per_page * (@current_page - OFFSET))
+           .limit(items_per_page)
+    end
+
+    def paginator(max_page_blocks: 7)
+      render partial: 'shared/paginator',
+             locals: { blocks: pages_to_display(@current_page,
+                                                max_page_blocks,
+                                                @last_page),
+                       current_page: @current_page,
+                       previous_link_css_class: previous_link_css_class,
+                       next_link_css_class: next_link_css_class,
+                       last_page: @last_page }
+    end
+
+    def page_item_css_class(page, current_page)
+      if page == current_page
+        'page-item active'
+      elsif page.to_s == ELLIPSES
+        'page-item disabled'
+      end
+    end
+
+    private
+
+    def previous_link_css_class
+      @current_page == FIRST_PAGE ? 'page-item disabled' : 'page-item'
+    end
+
+    def next_link_css_class
+      @current_page == @last_page ? 'page-item disabled' : 'page-item'
+    end
+
+    def pages_to_display(current_page, max_page_blocks, number_of_pages)
+      if all_pages_fit?(number_of_pages, max_page_blocks)
+        complete_pagination(number_of_pages)
+      elsif current_page_at_the_beginning?(current_page, max_page_blocks)
+        beginning_pagination(max_page_blocks)
+      elsif current_page_at_the_end?(current_page, number_of_pages,
+                                     max_page_blocks)
+        end_pagination(max_page_blocks, number_of_pages)
+      else
+        general_pagination(current_page, max_page_blocks)
+      end
+    end
+
+    def all_pages_fit?(number_of_pages, max_page_blocks)
+      number_of_pages < max_page_blocks
+    end
+
+    def current_page_at_the_beginning?(current_page, max_page_blocks)
+      current_page <= (max_page_blocks / DIVIDER)
+    end
+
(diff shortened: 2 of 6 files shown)


app/views/shared/_paginator.html.erb
```diff
@@ -0,0 +1,39 @@
...
+        link_to bets_path(page: current_page - 1), class: 'page-link' do
+          content_tag :i, nil, class: 'fa fa-angle-left'
+        end
+      end
+    %>
+
+    <% blocks.each do |page|%>
+      <%= render 'shared/blocks', page: page,
+        page_item_css: page_item_css(page,
+                                     current_page) %>
```
jacquesporveau: This is kind of weird code style. You've decided to put 2 arguments on the first line and one on the second.

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,106 @@
+module Utilities
+  module Paginator
+    def items_to_display(model:, items_per_page: 5, current_page: 1)
+      @items_per_page = items_per_page
+      @current_page = current_page.to_i
```
jacquesporveau: Why are we calling `to_i` on current page? I would assume we are passing in a integer to this method and not a float or string.

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,106 @@
...
+      @current_page == @last_page ? 'page-item disabled' : 'page-item'
+    end
+
+    def pages_to_display(current_page, max_page_blocks, number_of_pages)
+      if all_pages_fit?(number_of_pages, max_page_blocks)
+        complete_pagination(number_of_pages)
+      elsif current_page_at_the_beginning?(current_page, max_page_blocks)
+        beginning_pagination(max_page_blocks)
+      elsif current_page_at_the_end?(current_page,
+                                     number_of_pages, max_page_blocks)
```
jacquesporveau: One argument on the first line and two on the second.

spec/modules/paginator_spec.rb
```diff
@@ -0,0 +1,93 @@
+require 'rails_helper'
+
+RSpec.describe Utilities::Paginator, type: :module do
+  let!(:dummy) { Class.new { extend Utilities::Paginator } }
```
jacquesporveau: Perhaps this could be called 'paginator'?

lib/utilities/paginator.rb
ghost: Ah yeah, it's because in my BetsController I pass the parameter current_page: params[:page] as the argument. So, in this case it's either I apply it here just to be safe and make sure it's always an int or I can pass it there.
Please let me know what works best.

lib/utilities/paginator.rb
ghost: I remember why I did it now - Rubocop won't let me have 11 lines of code, so. I shuffled it around - 2 lines on top, one at the bottom, hope it looks better.

app/views/shared/_paginator.html.erb
ghost: Thanks, just noticed!

lib/utilities/paginator.rb
jacquesporveau: Maybe just one argument per line and then we can be consistent.

lib/utilities/paginator.rb
jacquesporveau: Yeah I'd like to see that done in the controller. To have current page referring to a string or integer depending on where it is being used is odd. Also is there an instance we need it to be a string? Usually the string "1" is pretty useless.

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,106 @@
...
+             locals: { blocks: pages_to_display(@current_page,
+                                                max_page_blocks,
+                                                @last_page),
+                       current_page: @current_page,
+                       previous_link_css: previous_link_css,
+                       next_link_css: next_link_css,
+                       last_page: @last_page }
+    end
+
+    def page_item_css(page, current_page)
```
natyork: Perhaps a more semantic name for this method would be something like ```page_item_class_name```?

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,106 @@
...
+      if page == current_page
+        'page-item active'
+      elsif page.to_s == '...'
```
natyork: You use the ellipses (```...```) in several places - it might be worth making it a constant, in case you ever want to change it in the future. This will also allow you to give it a more descriptive/semantic name so that you don't have 'Magic Numbers' (or text in this case) in your code.

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,106 @@
...
+        'page-item disabled'
+      end
+    end
+
+    private
+
+    def previous_link_css
+      @current_page == 1 ? 'page-item disabled' : 'page-item'
```
natyork: Magic number - this could be made a constant such as ```FIRST_PAGE``` instead of 1.

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,106 @@
...
+        general_pagination(current_page, max_page_blocks)
+      end
+    end
+
+    def all_pages_fit?(number_of_pages, max_page_blocks)
+      number_of_pages < max_page_blocks
+    end
+
+    def current_page_at_the_beginning?(current_page, max_page_blocks)
+      current_page <= (max_page_blocks / 2)
```
natyork: magic number

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,106 @@
...
+    end
+
+    def current_page_at_the_end?(current_page, number_of_pages, max_page_blocks)
+      current_page > (number_of_pages - (max_page_blocks / 2))
+    end
```
natyork: magic number

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,106 @@
...
+      if max_page_blocks == 1
+        [current_page]
+      elsif max_page_blocks == 2
+        [current_page, '...']
+      elsif max_page_blocks.odd?
+        odd_pages(current_page, max_page_blocks)
+      else
+        even_pages(current_page, max_page_blocks)
+      end
+    end
```
natyork, line 81: magic numbers

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    def beginning_pagination(max_page_blocks)
+      start = 1
+      finish = max_page_blocks - 1
+      [*(start..finish), '...']
+    end
```
natyork: magic numbers

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    def end_pagination(max_page_blocks, number_of_pages)
+      start = number_of_pages - max_page_blocks + 2
+      finish = number_of_pages
+      ['...', *(start..finish)]
+    end
```
natyork: magic number

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    def complete_pagination(last_page)
+      [*1..last_page]
+    end
```
natyork: magic number

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    def odd_pages(current_page, max_page_blocks)
+      start = (current_page - (max_page_blocks - 1) / 2) + 1
+      finish = (current_page + (max_page_blocks - 1) / 2) - 1
+      ['...', *(start..finish), '...']
+    end
```
natyork: magic numbers

lib/utilities/paginator.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    def even_pages(current_page, max_page_blocks)
+      start = (current_page - (max_page_blocks - 1) / 2) + 1
+      finish = (current_page + (max_page_blocks / 2)) - 1
+      ['...', *(start..finish), '...']
+    end
```
natyork: magic numbers

natyork (commented):
Overall looks pretty good, but I've flagged areas where you are using so-called 'magic numbers'. It would make your code much more readable and maintainable to use semantically-named constants rather than raw numbers/text that require interpretation throughout your code.

PR #62, opened 2017-03-28
Display all bets

ghost:
All bets are displayed on /bets page.


app/views/bets/index.html.erb
```diff
@@ -5,5 +5,9 @@
       <h4>This page will display all the existing bets (25/page) and allow the user to use custom search criteria.</h4>
     </div>
   </div>
-  <%= render 'bets/display_bets' %>
+  <div class="d-flex flex-column">
+    <div class="p-2">
+      <%= render 'display_bets', bets: @bets %>
```
fredericboivin: What is the intent behind `display_bets` partial ? (versus rendering the table in the index page and rendering individual bets through the `_bet` partial)

app/views/bets/_display_bets.html.erb
```diff
@@ -5,10 +5,16 @@
...
           <th class="col-md-2">Created at</th>
+          <th class="col-md-2">Expires at</th>
           <th class="col-md-2">Total amount bet</th>
-          <th class="col-md-6">Bet description</th>
+          <th class="col-md-4">Bet description</th>
         </tr>
       </thead>
+      <tbody>
+        <% bets.each do |bet| %>
+          <%= render 'bet', bet: bet %>
```
fredericboivin: You can replace this with `<%= render bet %>` as Rails will infer that the `_bet` partial can be used to render a Bet model. I don't feel strongly about one way or the other, just thought you should know in case you encounter it

app/views/bets/_bet.html.erb
```diff
@@ -0,0 +1,7 @@
+<tr>
+  <td class="col-md-2"><%= bet.creator_id %></td>
+  <td class="col-md-2"><%= bet.created_at %></td>
+  <td class="col-md-2"><%= bet.expires_at %></td>
+  <td class="col-md-2"><%= bet.bet_total %></td>
+  <td class="col-md-4"><%= bet.bet %></td>
```
fredericboivin, line 6: It might be out a scope for this pull request but I would recommend renaming this field as it is confusing. You refer to this column as "Bet Description" in the table header so maybe this should be called `description`.

qr8r: Probably worth doing at this point. You're the third person to point this out.
Would be a good time to learn the `sed` bash command

_EDIT_
ALthough, it should be done as a separate PR  I will write a store for it

qr8r:
@YarekTenko Cant review. Currently experiencing this locally 
https://www.pivotaltracker.com/story/show/142714857

qr8r (approved):
There is an N+1 Query when visiting this page. I will write a story to fix it, but I am happy to approve this comments

fredericboivin (approved):

PR #66, opened 2017-04-03
Fix date conversion problem when creating bet

ghost:
Pivotal: https://www.pivotaltracker.com/story/show/143166175
Added proper formatting to the jquery-datepicker so it adds user's time zone to the date string as well.

### DIFF
--- a/app/views/shared/_bet_display.html.erb
+++ b/app/views/shared/_bet_display.html.erb
@@ -12,7 +12,7 @@
         </div>
 
         <div class="card-footer text-danger">
-          Expires on: <%= bet.expires_at.strftime('%A, %B %d %Y, %l:%M %P') %>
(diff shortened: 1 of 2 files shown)


app/views/shared/_bet_display.html.erb
```diff
@@ -12,7 +12,7 @@
         </div>
 
         <div class="card-footer text-danger">
-          Expires on: <%= bet.expires_at.strftime('%A, %B %d %Y, %l:%M %P') %>
+          Expires on: <%= bet.expires_at.localtime.strftime('%A, %B %d %Y, %l:%M %P') %>
```
natyork, line 15: This is cool - I didn't know rails had a localtime method! Looks good :)

qr8r (approved):

PR #67, opened 2017-04-03
Rename bet column to description

ghost:
Pivotal: https://www.pivotaltracker.com/story/show/142709805
Renamed all the instances of "bet" that referred to bet text to "bet_description", because the bad naming caused confusion.

PR #68, opened 2017-04-03
Add seeds

ghost:
Pivotal: https://www.pivotaltracker.com/story/show/141706797
Dependent on #67 
Added seeds to quickly load functional test database.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,7 +1,109 @@
-# This file should contain all the record creation needed to seed the database with its default values.
-# The data can then be loaded with the rails db:seed command (or created alongside the database with db:setup).
-#
-# Examples:
-#
-#   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
-#   Character.create(name: 'Luke', movie: movies.first)
+bob = User.create(username: 'Bob', email: 'bob@mail.com', password: '123')
+sarah = User.create(username: 'Sarah', email: 'sarah@mail.com', password: '123')
+ivan = User.create(username: 'Ivan', email: 'ivan@mail.com', password: '123')
+maria = User.create(username: 'Maria', email: 'maria@mail.com', password: '123')
+
+# Resolved bets
+bet1 = Bet.create(bet_description: 'Ninjas vs Pirates',
+                  status: 0,
+                  creator_id: bob.id,
+                  expires_at: 1.year.from_now)
+b1_option1 = BetOption.create(option_text: 'Ninjas win',
+                              bet: bet1)
+b1_option2 = BetOption.create(option_text: 'Pirates win',
+                              bet: bet1)
+
+UserBet.create(user: bob,
+               bet: bet1,
+               bet_option: b1_option1,
+               amount_bet: 100)
+UserBet.create(user: sarah,
+               bet: bet1,
+               bet_option: b1_option2,
+               amount_bet: 215)
+UserBet.create(user: ivan,
+               bet: bet1,
+               bet_option: b1_option2,
+               amount_bet: 19.99)
+UserBet.create(user: maria,
+               bet: bet1,
+               bet_option: b1_option1,
+               amount_bet: 143.53)
+bet1.resolve(winning_option: b1_option2)
+
+bet2 = Bet.create(bet_description: 'Soccer: England VS Spain',
+                  creator_id: ivan.id,
+                  expires_at: 1.year.from_now,
+                  status: 0)
+b2_option1 = BetOption.create(option_text: 'England wins 2:0',
+                              bet: bet2)
+b2_option2 = BetOption.create(option_text: 'Spain wins 4:3',
+                              bet: bet2)
+b2_option3 = BetOption.create(option_text: 'England wins 1:0',
+                              bet: bet2)
+b2_option4 = BetOption.create(option_text: 'Draw',
+                              bet: bet2)
+
+UserBet.create(user: bob,
+               bet: bet2,
+               bet_option: b2_option1,
+               amount_bet: 25.34)
+UserBet.create(user: sarah,
+               bet: bet2,
+               bet_option: b2_option2,
+               amount_bet: 64.24)
+UserBet.create(user: ivan,
+               bet: bet2,
+               bet_option: b2_option3,
+               amount_bet: 34.35)
+UserBet.create(user: maria,
+               bet: bet2,
+               bet_option: b2_option4,
+               amount_bet: 56.99)
+bet2.resolve(winning_option: b2_option4)
+
+# Unresolved bets
+bet3 = Bet.create(bet_description: 'I bet Ivan will not deliver '\
+                  'the project on time',
+                  status: 1,
+                  expires_at: 1.year.from_now,
+                  creator_id: sarah.id)
+b3_option1 = BetOption.create(option_text: 'He will',
+                              bet: bet3)
+b3_option2 = BetOption.create(option_text: 'He won\'t',
+                              bet: bet3)
+UserBet.create(bet: bet3,
+               user: bob,
+               bet_option: b3_option2,
+               amount_bet: 13.5)
+UserBet.create(bet: bet3,
+               user: maria,
+               bet_option: b3_option1,
+               amount_bet: 20.99)
+
+# Expired bets
+bet4 = Bet.new
+Timecop.freeze(Date.new(2016, 8, 17)) do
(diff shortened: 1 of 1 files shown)


jacquesporveau:
I am seeing a lot more than just add seeds in here, you have a merge commit and merge conflicts. I would spend some time 'gitting' this ready for review.

ghost:
@jacquesporveau My bad, picked the wrong branch to rebase onto.

db/seeds.rb
```diff
@@ -1,7 +1,128 @@
...
 # Examples:
 #
-#   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
+#   movies = Movie.create([{ name: 'Star Wars' },
+#                          { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
+
+bob = User.create(username: 'Bob', email: 'bob@mail.com', password: '123')
+sarah = User.create(username: 'Sarah', email: 'sarah@mail.com', password: '321')
+ivan = User.create(username: 'Ivan', email: 'breku@mail.com', password: 'ur')
```
jacquesporveau: Just thinking if you ever want to sign in as these users it may be nice to have a high level of consistency with their emails and password. I notice all other Users emails follow a pattern besides this one. Also you could probably make all the passwords the same.

db/seeds.rb
```diff
@@ -1,7 +1,128 @@
...
+maria = User.create(username: 'Maria', email: 'maria@mail.com', password: 'm!a')
+
+# Resolved bets
+bet1 = Bet.create(bet_description: 'Ninjas vs Pirates',
```
jacquesporveau: If create fails due to validation is fails silently. use `create!` to get an error if creation fails. That was you know if your seed script is failing to do what you think it is doing.

db/seeds.rb
```diff
@@ -1,7 +1,128 @@
...
+
+# Unresolved bets
+bet3 = Bet.create(bet_description: 'I bet Ivan will not deliver '\
+                  'the project on time',
+                  status: 1,
+                  expires_at: DateTime.tomorrow,
+                  creator_id: sarah.id)
+b3_option1 = BetOption.create(option_text: 'He will',
+                              bet: bet3)
+b3_option2 = BetOption.create(option_text: 'He won\'t',
```
jacquesporveau: A preference thing but I like to avoid escaping characters in strings. Two ways you could do that would be 
`"He won't"`
`'He wont'`

Up to you. Maybe you are OK with escaping characters.

db/seeds.rb
```diff
@@ -1,7 +1,128 @@
...
+               amount_bet: 13.5)
+UserBet.create(bet: bet3,
+               user: maria,
+               bet_option: b3_option1,
+               amount_bet: 20.99)
+
+# Expired bets
+bet4 = Bet.new
+Timecop.freeze(Date.new(2016, 8, 17)) do
+  bet4 = Bet.create(bet_description: 'Rio Olympics: who will win?',
```
jacquesporveau, line 87: You are setting bet4 on line 96 to a new instance of `Bet` and then rewriting it on line 98.

db/seeds.rb
```diff
@@ -1,7 +1,128 @@
...
+               bet_option: b3_option2,
+               amount_bet: 13.5)
+UserBet.create(bet: bet3,
+               user: maria,
+               bet_option: b3_option1,
+               amount_bet: 20.99)
+
+# Expired bets
+bet4 = Bet.new
+Timecop.freeze(Date.new(2016, 8, 17)) do
```
jacquesporveau, line 86: Why are we freezing time?

db/seeds.rb
```diff
@@ -1,7 +1,128 @@
...
+               user: maria,
+               bet_option: b4_option2,
+               amount_bet: 53.3)
+UserBet.create(bet: bet4,
+               user: ivan,
+               bet_option: b4_option3,
+               amount_bet: 35.2)
+
+# Filler bets
+50.times do |bet|
```
jacquesporveau: I'd like this renamed to `i` representing an index. An integer is not a bet.

db/seeds.rb
```diff
@@ -1,7 +1,128 @@
...
+  Bet.create(bet_description: "Filler test bet #{bet + 1}",
+             expires_at: DateTime.tomorrow,
+             creator_id: rand(User.count) + 1,
```
jacquesporveau: This could also be `rand(User.count + 1)`.

db/seeds.rb
```diff
@@ -1,7 +1,128 @@
-# This file should contain all the record creation needed to seed the database with its default values.
-# The data can then be loaded with the rails db:seed command (or created alongside the database with db:setup).
+# This file should contain all the record creation needed to seed the database
```
jacquesporveau: Were these changes because of rubocop? If so we can probably just disable it from picking up comments as there are tonnes of rails files with long comments that we don't want to have to change every time we edit them.

You could also just delete these comments. You know what goes here and why.

jacquesporveau (changes requested):

.gitignore
```diff
@@ -20,3 +20,6 @@
 # Ignore Byebug command history file.
 .byebug_history
 coverage/
+
+# Ignore vendor/cache/ files.
+/vendor/cache
```
natyork: I am curious why you are updating the gitignore in this PR? Should this belong elsewhere, or it's own separate PR?

db/seeds.rb
natyork: Agree with Dennis - for staff tracker rubocop we exclude all files in the db folder.

db/seeds.rb
```diff
@@ -1,7 +1,128 @@
...
+               amount_bet: 53.3)
+UserBet.create(bet: bet4,
+               user: ivan,
+               bet_option: b4_option3,
+               amount_bet: 35.2)
+
+# Filler bets
+50.times do |bet|
+  Bet.create(bet_description: "Filler test bet #{bet + 1}",
+             expires_at: DateTime.tomorrow,
```
natyork: Does this work the way you intend? I would have thought that this would only update the date at the time you run the seed - if you need unexpired bets in your database for development purposes, would it be more useful to create an expiry date in the far future?

db/seeds.rb
ghost: Good idea, thanks!

db/seeds.rb
ghost: Sure, will do.

db/seeds.rb
ghost: Ah, gotcha, didn't know that!

db/seeds.rb
ghost: I want to create an expired bed for testing purposes, with the way the validator is set up right now, it won't allow me to create bet with an already expired date, so I had to do this, hope it makes it more clear.

db/seeds.rb
ghost: Yeah, I just declared it so that I could access it out of the block after changing it, the Bet.new was chosen arbitrarily.

db/seeds.rb
ghost: Sure!

.gitignore
ghost: Sure, will move it out, thanks!

db/seeds.rb
ghost: I could add a couple of years to it, but ultimately doesn't matter to me because I tend to drop-migrate the database a lot. Thanks for the suggestion!

db/seeds.rb
jacquesporveau: You could just set it to `1.hour.from_now`

db/seeds.rb
```diff
@@ -1,7 +1,117 @@
...
+               user: maria,
+               bet_option: b4_option2,
+               amount_bet: 53.3)
+UserBet.create(bet: bet4,
+               user: ivan,
+               bet_option: b4_option3,
+               amount_bet: 35.2)
+
+# Filler bets
+50.times do |index|
```
qr8r: Can we kill these guys? I think they just create noise. With seeds I only want enough data to get the site working.

qr8r (approved):

qr8r (changes requested):
Accidental approval

db/seeds.rb
ghost: Something new every day, thanks!

qr8r (approved):

jacquesporveau (approved):

PR #36, opened 2017-03-10
Users can join bets

ghost:
Pivotal Links: 
https://www.pivotaltracker.com/story/show/140512067
https://www.pivotaltracker.com/story/show/141507823

Dependent on #35  
Each BetOption now has a "Bet!" button, which takes a user to the page where they can enter their betting amount and place it or cancel and return to the Bet page.

PR #70, opened 2017-04-06
Fix date conversion problem when creating bet

ghost:
Re-upload: accidentally merged with #36 
Pivotal: https://www.pivotaltracker.com/story/show/143166175
Added proper formatting to the jquery-datepicker so it adds user's
timezone to the date string as well.

### DIFF
--- a/app/assets/javascripts/lib/date_picker.js
+++ b/app/assets/javascripts/lib/date_picker.js
@@ -1,5 +1,6 @@
 $(document).on("turbolinks:load", function(){
   $(".jquery-datepicker").datetimepicker({
-    minDate: "0"
+    minDate: "0",
+    format: "Y-m-d G:i:s P"
   });
 });
--- a/app/views/shared/_bet_display.html.erb
+++ b/app/views/shared/_bet_display.html.erb
@@ -12,7 +12,7 @@
         </div>
 
         <div class="card-footer text-danger">
-          Expires on: <%= bet.expires_at.strftime('%A, %B %d %Y, %l:%M %P') %>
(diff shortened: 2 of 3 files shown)


app/controllers/user_bets_controller.rb
```diff
@@ -16,4 +18,36 @@ def user_bet_params
     params.require(:user_bet)
           .permit(:user_id, :bet_id, :amount_bet, :bet_option_id)
   end
+
+  def verify_fields
```
qr8r: The controller is not the correct place to perform validations

app/controllers/user_bets_controller.rb
```diff
@@ -1,4 +1,6 @@
 class UserBetsController < ApplicationController
+  before_action :verify_fields, only: :create
```
qr8r: Im not sure this commit fits with this PR.

app/assets/javascripts/lib/date_picker.js
```diff
@@ -1,5 +1,6 @@
 $(document).on("turbolinks:load", function(){
   $(".jquery-datepicker").datetimepicker({
-    minDate: "0"
+    minDate: "0",
+    format: "Y-m-d G:i:s P"
```
qr8r, line 4: Perhapse for now we might just want to set the rails time zone to be pacific https://www.varvet.com/blog/working-with-time-zones-in-ruby-on-rails/

qr8r (changes requested):

app/controllers/user_bets_controller.rb
ghost: Yeah, was redone in the previous PR, thank you for catching this one.

qr8r (approved):

